PR #2, opened 2015-02-19
PIRANHA-3146 append instance uuid into inspect output

13W:


trentm:
replied on email discussion

PR #9, opened 2015-03-06
DOCKER-228 SDC-Docker API crashes when creating container without HostCo...

13W:
...nfig param


joshwilsdon:
Fixed in f15421ae9d0187d4cfdce12f965caaf20551a2f9

PR #10, opened 2015-03-11
DOCKER-145 'docker history ID' fails, works with an image *name*

13W:


joshwilsdon:
fixed with e477c53807fcba50b3b75d63b01677b445a26779

PR #11, opened 2015-03-16
Docker changes options for SDC-Docker container logs

yankeys:


trentm:
I'm leery of the `== false` (double-equals and not triple-equals).

Presumably this is to support "--follow" for https://docs.docker.com/reference/commandline/cli/#logs `docker logs`?  Does this make it work with sdc-docker?

PR #12, opened 2015-03-17
correct handling of logs' --follow flag needed for PIRANHA-3246

dan-luk:


trentm:
This should be fixed in commit d09b744. Thanks.

dan-luk:
thanks Trent

PR #21, opened 2015-03-26
DOCKER-271 more fixes for deleting image tags

dan-luk:


cachafla:
Patch landed here: https://github.com/joyent/sdc-docker/commit/f85db77928623ad0cf2cd3326902851abba2f138

trentm:
@dan-luk Please let's not re-used DOCKER-271 for these :)

PR #23, opened 2015-04-10
In non-overlay network mode, allow selecting which network is "external"

arekinath:
In particular, don't enforce that "external" must be on a nic_tag named "external". This makes sdc-docker easier to try out in pre-existing SDC deployments that don't have overlays enabled yet.

This can take in a metadata option, EXTERNAL_NET to choose a name other than "external" to look for when selecting the network for docker containers.

### DIFF
--- a/lib/backends/sdc/containers.js
+++ b/lib/backends/sdc/containers.js
@@ -163,7 +163,8 @@ function addNetworksToPayload(opts, container, payload, callback) {
     // No fabrics configured - fall back to provisioning on the external
     // network
 
-    listNetworks(opts, {name: 'external'}, function (err, networks) {
+    var netName = opts.config.externalNetwork || 'external';
+    listNetworks(opts, {name: netName}, function (err, networks) {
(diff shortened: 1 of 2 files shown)


lib/backends/sdc/containers.js
```diff
@@ -163,7 +163,8 @@ function addNetworksToPayload(opts, container, payload, callback) {
     // No fabrics configured - fall back to provisioning on the external
     // network
 
-    listNetworks(opts, {name: 'external'}, function (err, networks) {
+    var netName = opts.config.externalNetwork || "external";
```
trentm: Thanks. Did 'make check' complain about the double-quoted string?

arekinath:
Ah, sorry. Corrected to single quotes. I made the change initially on the actual docker zone here and forgot to keep the original to diff, so I put it back together by re-typing because it was so short. The double quotes snuck in on the way and I didn't re-run `make check`.

PR #25, opened 2015-04-15
Added dns run parameters as unimplemented features

dekobon:
This is only a change to the documentation to indicate that this is an unimplemented feature.


lloydde:
Feature already committed.
@dekobon keep the pull requests coming! :beer:

PR #27, opened 2015-04-22
OpenSSH_6.8 changed the output of "ssh-keygen -l"

lloydde:
Fixes DOCKER-324.

### DIFF
--- a/tools/sdc-docker-setup.sh
+++ b/tools/sdc-docker-setup.sh
@@ -109,6 +109,22 @@ function dockerInfo
 }
 
 
+
+# Return ssh fingerprint in the form "he:xh:ex:he:xh:..."
+# OpenSSH_6.8 changes the "-l" flag output.
+function sshGetMD5Fingerprint() {
+    local sshPubKeyPath=$1
+    local s
+    s=$(ssh-keygen -E md5 -l -f "$sshPubKeyPath" 2> /dev/null)
+    if [[ $? -eq  0 ]]; then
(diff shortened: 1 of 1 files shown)


tools/sdc-docker-setup.sh
```diff
@@ -109,6 +109,21 @@ function dockerInfo
 }
 
 
+# Return ssh fingerprint in the form "he:xh:ex:he:xh..."
+# OpenSSH_6.8 changes the -l flag output.
+function sshGetMD5Fingerprint() {
+    local s
+    sshPubKeyPath=$1
```
trentm: should use `local` too I believe

tools/sdc-docker-setup.sh
```diff
@@ -109,6 +109,21 @@ function dockerInfo
...
+    s=$(ssh-keygen -E md5 -l -f "$sshPubKeyPath" 2> /dev/null)
+    if [[ $? -eq  0 ]]; then
+        echo "$s" | awk '{print $2}' | tr -d '\n' | cut -c 5-;
```
trentm: is `| cut -c 5-;` accurate? Doesn't it use a ':' delimiter?  If so, would prefer `cut -d: -f2-`. Fine either way tho. I'm being pedantic. :)

trentm:
otherwise, lgtm!

lloydde:
Merged in 729e5898ffa91bd14736e192529a467fce7bef31.

PR #30, opened 2015-05-06
DOCKER-335 partial

misterbisson:
Doesn't attempt to address the TLS cert issue, as I'm not clear where we'd want to put the ca.pem here.

Affects DOCKER-335
See also DOCKER-275


twhiteman:
This change is incorporated by the DOCKER-275 changes.

See the https://github.com/joyent/sdc-docker/blob/master/README.md describes the TLS cert updating process, and you'll need to re-run the *tools/sdc-docker-setup.sh* in order to properly setup the ca.pem file.

PR #31, opened 2015-05-13
Scrubbed related APIs and typos in the feature section for Docker API docs

askfongjojo:
This is the first batch on changes on the more straight-forward edits and cross-references. Expanding on the features (networking, resource allocation, etc) will come next.

### DIFF
--- a/docs/api/commands/events.md
+++ b/docs/api/commands/events.md
@@ -6,4 +6,4 @@ This command is currently unimplemented. Follow [DOCKER-78](http://smartos.org/b
 
 ## Related
 
-- [`MachineAudit`](https://apidocs.joyent.com/cloudapi/#MachineAudit) and `GET /my/machines/:id/audit` in CloudAPI
+- [`sdc-getmachineaudit`](https://apidocs.joyent.com/cloudapi/#MachineAudit) and `GET /my/machines/:id/audit` in CloudAPI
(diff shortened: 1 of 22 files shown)


docs/api/commands/events.md
```diff
@@ -6,4 +6,4 @@ This command is currently unimplemented. Follow [DOCKER-78](http://smartos.org/b
 
 ## Related
 
-- Insert a list of related Docker and CloudAPI methods here
\ No newline at end of file
+- [`MachineAudit`](https://apidocs.joyent.com/cloudapi/#MachineAudit) and `GET /my/machines/:id/audit` in CloudAPI
```
misterbisson: Should we consistently use the CLI names for CloudAPI since that's what we're doing for the Docker commands?

This could be `sdc-getmachineaudit`, I think.

docs/api/features/resources.md
```diff
@@ -1,8 +1,9 @@
-# Resource selection
+# Resource Allocation
```
misterbisson: Headings should be sentence case.

misterbisson:
Really good changes. Thanks for this huge work.

docs/api/features/resources.md
askfongjojo: Oh ok. I was wondering about this too and followed "Restart Policies" as an example. Should I change that to "Restart policies"?

docs/api/commands/events.md
askfongjojo: I was under the impression that there was no CLI for this one when looking at the CloudAPI docs. I'll fix this.

PR #35, opened 2015-05-28
doc: exec no privileged option.

lloydde:
https://github.com/docker/docker/blob/master/docs/sources/reference/commandline/cli.md#exec
Also:
- textwidth=79

### DIFF
--- a/docs/api/commands/exec.md
+++ b/docs/api/commands/exec.md
@@ -6,16 +6,17 @@
 
       -d, --detach=false         Detached mode: run command in the background
       -i, --interactive=false    Keep STDIN open even if not attached
-      --privileged=false         Give extended privileges to the command
       -t, --tty=false            Allocate a pseudo-TTY
       -u, --user=                Username or UID (format: <name|uid>[:<group|gid>])
 
 The `docker exec` command runs a new command in a running container.
 
 The command started using `docker exec` only runs while the container's primary
-process (`PID 1`) is running, and it is not restarted if the container is restarted.
+process (`PID 1`) is running, and it is not restarted if the container is
+restarted.
 
-If the container is paused, then the `docker exec` command will fail with an error:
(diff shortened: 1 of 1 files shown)


docs/api/commands/exec.md
```diff
@@ -44,7 +45,9 @@ This will create a new Bash session in the container `ubuntu_bash`.
 
 ## Divergence
 
-There is no known divergence between the Triton SDC Docker and Docker Inc. implementations of this method. Please contact Joyent support or file a ticket if you discover any.
+There is no known divergence between the Triton SDC Docker and Docker Inc.
```
misterbisson: This is not correct if `--privileged` is unsupported.

docs/api/commands/exec.md
```diff
@@ -44,7 +45,9 @@ This will create a new Bash session in the container `ubuntu_bash`.
...
+implementations of this method. Please contact Joyent support or file a ticket
```
misterbisson: Please don't hard-wrap english language content. It breaks the semantics of the language.

jclulow: It's actually just fine to carry on hard-wrapping _Markdown Source_, which is what this file is.  The semantic construct of a paragraph, in Markdown, is a contiguous set of lines separated by a blank line.  This is essentially a code style thing, and is generally a per-repository decision.

docs/api/commands/exec.md
lloydde: Sorry, I'm a little understanding. "--privileged" has been removed from the upstream documentation for this command, see https://github.com/docker/docker/blob/master/docs/sources/reference/commandline/cli.md#exec

docs/api/commands/exec.md
lloydde: @misterbisson you want this one paragraph formatted differently than the rest of the file? I'd highly recommend against changing the format from upstream.

PR #37, opened 2015-06-08
Convert api docs to Kirby CMS format

fitzage:


trentm:
WOn't take this. We're processing the sdc-docker docs (leaving the format in this repo as is) to a separate repo for presentation.

PR #40, opened 2015-06-18
func sshGetMD5Fingerprint wasn't working when invoked 'sh', fixes #39.

### LINKED ISSUE #39: sdc-docker-setup.sh script does not work on OS X 10.10
```console
$ sh sdc-docker-setup.sh https://us-east-1.api.joyentcloud.com myuser ~/.ssh/id_rsa 
Setting up Docker client for SDC using:
    CloudAPI:        https://us-east-1.api.joyentcloud.com
    Account:         myuser
    Key:             /Users/epoch/.ssh/id_rsa

$ echo $?
1
```

Digging into the script, it looks like the `sshGetMD5Fingerprint` function is at fault, specifically this command:

```sh
    s=$(ssh-keygen -E md5 -l -f "$sshPubKeyPath" 2> /dev/null)
```

The version of ssh-keygen that I have on my Mac doesn't have a `-E` option.

```console
$ ssh -V
OpenSSH_6.2p2, OSSLShim 0.9.8r 8 Dec 2011
```

Based on the comment above the `sshGetMD5Fingerprint` function, it looks like this was written assuming OpenSSH 6.8 or above. This is going to be a problem for other OS X users with the default SSH client...

lloydde:


lloydde:
Not yet tested on recent version of OpenSSH.

rmustacc:
This isn't an sh script though, why are we invoking it as sh? If we want it to be a proper sh script, there's a lot more that's wrong with it.

lloydde:
That is an excellent question @rmustacc. @trentm thoughts?

I've confirmed that this change works when invoked with 'sh sdc-docker-setup.sh' on:
- Mac 10.10.3: GNU bash, version 3.2.57(1) / OpenSSL 0.9.8zd 8 Jan 2015
- nfnty/arch-mini: GNU bash, version 4.3.39(1) / openssh 6.8p1-3

rmustacc:
Sure, it works a bit when sh is actually bash. Unfortunately sh is not always bash (see, eg. smartos, debian, *BSD, etc.). An analysis as to why this change works is necessary and likely will need to be in a comment, as moving the assignment into the if in shell scripting is not generally done stylistically.

lloydde:
My first choice would be to update docs to invoke with 'bash' and kill this pull request.

Other than the if hack, I haven't found another way for the the invalid option from `ssh-keygen -E md5 -l -f "$sshPubKeyPath" 2> /dev/null` not to exit the function and result in the empty assignment `sshKeyId=`.

I haven't explored trapping, but my strong preference will be to update docs to invoke with `bash`.

trentm:
+1 to changing docs to say invoke with bash
If we want to support with just vanilla `sh`, then that should be separate work (with test cases).

PR #43, opened 2015-07-11
DOCKER-458: Can't specify memory for containers with 1.7 client

misterdjules:
Instead of getting the value `Memory` from just the root of the container
object, which is where docker clients < 1.7 put it, we also check in
`container.HostConfig.Memory`, which is where docker clients >= 1.7 put
it.

This commit adds integration tests for this change. These tests create
docker containers with a memory value different than the default memory
value and mimic the behavior of docker clients pre and post 1.7.

This commit also moves `loadConfigSync` to its own module so that it can
be used by tests that need to load the configuration to lookup default
values (such as `defaultMemory` in this case).

### DIFF
--- a/lib/backends/sdc/containers.js
+++ b/lib/backends/sdc/containers.js
@@ -927,14 +927,18 @@ function getPackage(opts, container, callback) {
     }, function (err, pkgs, count) {
             var candidate = {};
             var constraints = {};
-            var memory;
+            // Starting with docker client version 1.7, the memory
+            // for the host is passed in the container.HostConfig object,
+            // not directly in the container object.
+            var hostConfig = container.HostConfig || {};
+            var memory = Number(container.Memory || hostConfig.Memory);
 
-            if (container.Memory) {
+            if (memory) {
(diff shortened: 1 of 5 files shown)


misterdjules:
@joshwilsdon Updated the changes according to your code review, thank you :)

joshwilsdon:
I notice there's still MEMORY_IN_MBS while BYTES_IN_MBS was changed to BYTES_IN_MB, should we change MEMORY_IN_MBS too?

Otherwise: LGTM

misterdjules:
@joshwilsdon I wrote `MEMORY_IN_MBS` as "This value represents megabytes of data", hence the plural form, whereas `BYTES_IN_MBS` was changed to `BYTES_IN_MB` since it means "How many bytes in one megabyte`. However my English is not that great, so if that's not correct I'll change it :)

misterdjules:
Landed in c0cc2ee5c4bb77416c15f8a29bf60e37f83760d3. Thank you @joshwilsdon :+1:

PR #45, opened 2015-07-16
DOCKER-460: docker build should error out earlier if possible

misterdjules:
Separate implemented from unimplemented endpoints. Register
unimplemented ones before registering common handlers that imply
significant processing, such as (but not only) parsing requests' body.

For instance, this solve the problem when docker build would make the
server process the whole request's body before responding that the build
endpoint is not implemented.

This change also changes the behavior of the bodyParser handler so that
it errors if a request for an implemented handler passes through it.
This helps ensuring that further changes to the order in which handlers
are executed will still make unimplemented endpoints at least not go
through the body parser.

No regression in integration tests, and fixes the specific use case described in DOCKER-460.

/cc @trentm @twhiteman @joshwilsdon


orlandov:
LGTM

misterdjules:
@trentm Updated this PR according to our discussion during a previous code review.

trentm:
Discussed offline. LGTM.

misterdjules:
Merged with 4b869e24cb5c850e43cdf9f025d145abf1fe8c5d, thank you @trentm and @orlandov :+1:

PR #47, opened 2015-07-20
DOCKER-483: Audit log causes out of memory crashes

misterdjules:
This change prevents logging the body of a given request if its size is
superior to a given constant.

This prevents some of the out of memory crashes that are due to
stringifying large requests' bodies before logging them.

### DIFF
--- a/lib/audit-logger.js
+++ b/lib/audit-logger.js
@@ -1,0 +1,161 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+/*
+ * Copyright (c) 2015, Joyent, Inc.
+ */
+
+var assert = require('assert-plus');
+var bunyan = require('bunyan');
+
+var HttpError = require('restify').HttpError;
+
+// Default maximum length for requests/responses body that are logged
+// is 10 KBs. This can be overriden with the option object passed
+// to auditLogger.
+var DEFAULT_MAX_LOG_BODY_LENGTH = 10 * 1024;
+
+function getBodyToLog(body, options) {
+    var loggedBody;
+    options = options || {};
+
+    assert.object(options, 'options');
+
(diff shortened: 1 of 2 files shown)


misterdjules:
/cc @trentm @joshwilsdon @twhiteman  Looking for preliminary feedback.

There are other things we could do regarding this:

1. For requests with large payloads, we could log a portion of the body instead of not logging it at all, which might help us troubleshoot issues sometimes.

2. The threshold could also be made configurable in `config.json`.

misterdjules:
Mostly as a note to myself, after some discussions with @trentm and @joshwilsdon, here's what we'll do:

1. Create our own fork of the resitfy audit logs plugin.
2. Make that fork accept different configurations/values for requests and responses.
3. Make that fork not log buffers and truncate large strings (> 1 MB) in requests and responses' bodies by default.
4. Make that fork accept overrides for the threshold of what is considered to be too large for a request/response body.

misterdjules:
@trentm Updated this PR according to our discussion with a second commit. Could you please take a look and let me know what you think?

lib/audit-logger.js
```diff
@@ -0,0 +1,161 @@
...
+ * so that options can be set separately for the response body or request body
+ * respectively.
+ *
+ * - Each of these options sub-objects support the following properties:
+ *
+ *  * log: if true, the request or the response's body will be logged.
+ *  * buffers: if true, requests/responses' bodies that are buffers will
+ *    be logged. False by default.
+ *  * maxBodyLength: allow to specify the maximum size of body's data that
+ *    will be logged. DEFAULT_MAX_LOG_BODY_LENGTH (currently 10KBs) by default.
```
trentm: Just `maxLength`?  The "Body" is redundant.

trentm:
LGTM

lib/audit-logger.js
misterdjules: Indeed, good catch, thank you :+1: Fixed and updated this PR.

misterdjules:
Merged with 90fb9aa688d152fe6b1bbb124baa54c992ca762d.

PR #51, opened 2015-08-12
DOCKER-467: create config settings for server and api version

twhiteman:
Summary of changes:
* new *etc/defaults.json* file
  * not sure why another config file - why not use the sapi manifest file?
* docker server API version (1.19) and docker server version (1.7) moved to *etc/defaults.json* file
* sdc-docker app config first loads the default config (*etc/defaults.json*), then loads the sapi config (*etc/config.json*) and then merges these two config settings (sapi wins over default) to create the app.config
* update various places/tests where server/api version is used

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,8 @@ and officially supported version.
 Devs: When updating the sdc-docker server official version, you'll need to
 be sure to update the following:
 
-1. update SERVER_VERSION version in lib/common.js
-2. update the test client version in test/integration/helpers.js
-3. update the docker cli test client version in
+1. update both *API_VERSION* and *SERVER_VERSION* version in lib/constants.js
+2. update the docker cli test client version in
    globe-theatre/bin/nightly-test-docker-integration-cli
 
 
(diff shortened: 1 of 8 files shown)


lib/config-loader.js
```diff
@@ -11,15 +11,22 @@
...
+    var defaultsPath = path.resolve(__dirname, '..', 'etc', 'defaults.json');
+    opts.log.info('Loading default config from "%s"', defaultsPath);
+    var defaultConfig = JSON.parse(fs.readFileSync(defaultsPath, 'utf-8'));
+
     var configPath = path.resolve(__dirname, '..', 'etc', 'config.json');
     opts.log.info('Loading config from "%s"', configPath);
     var config = JSON.parse(fs.readFileSync(configPath, 'utf-8'));
 
+    config = common.mergeObjects(config, defaultConfig);
+
```
trentm: could just use the existing `common.objCopy` right?

```
config = /* parse defaultsPath */
config = common.objCopy(/* parse configPath */, config);
```

trentm:
@twhiteman is etc/defaults.json in this PR?

PR #60, opened 2016-01-13
DOCKER-633: add tests for copy on stopped containers

orlandov:


test/integration/copy.test.js
```diff
@@ -1,7 +1,7 @@
 /*
  * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * License, v. 2.0. If a copy of the MPL was not distripbuted with this
+ * file, You can obtain one at http://mozilla.org/MPL/3.0/.
```
notmatt: Typo - 'distripbuted' and MPL 3.0 (should be 2.0, there is no 3.0).

test/integration/copy.test.js
```diff
@@ -68,92 +73,302 @@ test('test initialization', function (tt) {
...
-
-        function onDocker(err, stdout, stderr) {
-            t.ifErr(err);
-            console.log(stdout.trim());
-            t.end();
+    vasync.waterfall([
+        function (next) {
+            createCopyOutFile(tt, ffn, nginxName, function (err, sha1) {
+                tt.ifErr(err, 'creating copy out file');
+                hash = sha1;
```
notmatt: `hash` is declared at the start of the test function, but only used within this scope, could use `sha1` directly in the `test.comment` immediately following. Should it perhaps be tested against the hash of the source file?

PR #62, opened 2016-01-21
DOCKER-502 support the use of a sdc package for creating containers

joshwilsdon:
This adds support for:

```
docker run --label com.joyent.package=<pkg> ...
```

where &lt;pkg&gt; can be:

 * a package name
 * a package UUID
 * the first 8 characters of the package UUID

it also adds support for:

```
docker ps --filter "label=com.joyent.package=<pkg>"
```

with the same set of possible values for &lt;pkg&gt;.


joshwilsdon:
I've addressed feedback from @askfongjojo and @trentm with 82518d056ec1b5dde85bcfa81f9d09b2b3d98535, specifically:

- added comments about why com.joyent.package is special both to docs and code
- added package label to inspect output
- documented order of evaluation
- fixed `+ '(name=' + opts.specifiedPackage +'))';`
- created DOCKER-677 to address package exclusion problems
- matched trent-style indentation
- included PACKAGE_SELECTION_LABEL in error messages for invalid package

what remains at this point is addressing any additional feedback and adding some tests.

lib/backends/sdc/containers.js
```diff
@@ -981,130 +999,387 @@ function renameLinks(opts, newName, callback) {
...
-                return;
+    var constraint = {};
+    var foundUuid;
+    var log = opts.log;
+    var nameMatches = [];
+    var uuidMatches = [];
+
+    constraint[PACKAGE_SELECTION_LABEL] = specifiedPackage;
+
+    if (pkgs.length === 1) {
```
misterdjules: It seems it would be more consistent to check that in the case where `pkgs.length === 1`, `pkgs[0]` actually matches `specificiedPackage`. Thus, it might be more consistent to not treat that as a special case.

lib/backends/sdc/containers.js
```diff
@@ -981,130 +999,387 @@ function renameLinks(opts, newName, callback) {
...
+    assert.string(opts.packagePrefix, 'opts.packagePrefix');
+    assert.uuid(opts.ownerUuid, 'opts.ownerUuid');
+    assert.optionalString(opts.specifiedPackage, 'opts.specifiedPackage');
+
+    var filter;
+
+    if (opts.specifiedPackage) {
+        if (common.isUUID(opts.specifiedPackage)) {
+            // if specifiedPackage is a UUID, we'll just look that up directly.
+            filter = '(&(active=true)'
```
misterdjules: We might want to use [ldap-filter](https://github.com/pfmooney/node-ldap-filter) to build these filters in the future, although I'm not sure how to represent the `(!(owner_uuids=*))` query with that module.

lib/backends/sdc/containers.js
```diff
@@ -1867,72 +2180,181 @@ function getContainerCount(opts, callback) {
...
+                });
+            }
+
+            if (invalidPackageName) {
+                cb(new errors.DockerError('invalid value for '
+                    + PACKAGE_SELECTION_LABEL + ': ' + invalidPackageName));
+                return;
+            }
+
+            loadPackages({
```
misterdjules: Is loading all packages needed e.g if no package selection label is specified?

lib/backends/sdc/containers.js
jclulow: @misterdjules presumably with [NotFilter](http://ldapjs.org/filters.html#notfilter).

misterdjules: @jclulow Absolutely, what I wasn't 100% sure of is not about translating the `!` operator, but more if `ldap-filter` was doing anything special with the `*` pattern. In other words, what I meant is: "please check that using ldap-filter's filters does a direct translation and what we expect, because I haven't checked that myself, so I'm not 100% sure".

lib/backends/sdc/containers.js
```diff
@@ -2017,15 +2476,35 @@ function getContainers(opts, callback) {
...
+            var objects = stash.objects;
+
+            if (opts.limit > 0) {
+                objects = objects.slice(0, opts.limit);
+            }
+
+            // last cb in chain, so we pass objects to results
+            cb(null, objects);
         }
+    ]}, function _pipelineCb(p_err, p_results) {
```
misterdjules: Are `p_err` and `p_results` prefixed with `p_` to avoid shadowing other variables named `err` and `results`? If so, is there some refactoring that wouldn't be too time consuming that would allow us to not do that? Otherwise my concern is that it's a missed opportunity to make the code not too confusing.

test/integration/helpers.js
```diff
@@ -1110,6 +1111,21 @@ function createSapiClient(callback) {
 
 
 /**
+ * Get a simple restify JSON client to PAPI.
+ */
+function createPapiClient(callback) {
+    createClientOpts('papi', function (err, opts) {
```
misterdjules: Nitpicking, but `assert.func(callback, 'callback`)` could be useful.

test/integration/packagelabels.test.js
```diff
@@ -0,0 +1,468 @@
...
+            name: '_bacon', // leading underscore invalid
+            errPrefix: invalidPrefix
+        }, {
+            name: 'bacon-', // trailing '-' is invalid
+            errPrefix: invalidPrefix
+        }, {
+            name: 'bacon!', // '!' is invalid
+            errPrefix: invalidPrefix
+        }, {
+            name: '💩', // nope
```
misterdjules: Is it that non-ASCII prefixes are not allowed?

misterdjules:
@joshwilsdon Left a few comments :)

lib/backends/sdc/containers.js
joshwilsdon: Very good point. Thanks!

lib/backends/sdc/containers.js
joshwilsdon: Originally it was not, but because of the fact that /containers/json includes the Labels for all packages, we need the map of uuid -> package name later on around line 2450. I realize that loading from PAPI each time adds some more complexity for now, but I've created DOCKER-687 and added a note about that above in loadPackages() so that we at least add caching of this data. The package data should be pretty static so loading it periodically even might be enough.

lib/backends/sdc/containers.js
joshwilsdon: Yes, the p_ prefixes are to avoid overshadowing the variables in the outer scope. I'll take a look here and see what I can do.

test/integration/helpers.js
joshwilsdon: Yep, good catch! Unfortunately this file is missing a lot of asserts. I'll have to add those in too.

test/integration/packagelabels.test.js
joshwilsdon: yep, I'll improve the comment.

joshwilsdon:
@misterdjules I've pushed 3445743eb24fbc8087962b7d57d57e611704279a which I think addresses all your comments except the one about ldap-filter. And for that one I created DOCKER-688. Let me know if you think there's additional work that should be done here or one of my changes seems incorrect.

lib/backends/sdc/containers.js
misterdjules: > because of the fact that /containers/json includes the Labels for all packages

Do we need that so that users get package names rather than UUIDs if they created their container(s) with a package name, and thus to have a better user experience?

misterdjules:
I left [an additional question](https://github.com/joyent/sdc-docker/pull/62#discussion_r51595742), but mostly for my own curiosity. Otherwise LGTM.

PR #70, opened 2016-03-07
DOCKER-728: read/write "triton.*" labels from/to VM.tags["triton.*"]

trentm:
DOCKER-721: Docker 1.10 ps/inspect errors out when label contains boolean values

### DIFF
--- a/lib/backends/sdc/containers.js
+++ b/lib/backends/sdc/containers.js
@@ -32,6 +32,7 @@ var common = require('../../../lib/common');
 var errors = require('../../../lib/errors');
 var images = require('./images');
 var Link = require('../../models/link');
+var triton_tags = require('../../triton-tags');
 var utils = require('./utils');
 var validate = require('../../validate');
 
@@ -225,11 +226,7 @@ function addFabricNetworksToPayload(opts, container, payload, callback) {
     }
 
     opts.app.ufds.getDcLocalConfig(opts.account.uuid, dc, function (err, conf) {
-        log.debug('conf:', conf);
-        log.debug({
-            err: err,
-            conf: conf,
-            account: opts.account.uuid},
+        log.debug({err: err, conf: conf, account: opts.account.uuid},
             'get DC local config');
 
         if (err || !conf || !conf.defaultnetwork) {
@@ -1475,6 +1472,9 @@ function buildVmPayload(opts, container, callback) {
     };
     var restartPolicy;
     var volumesFrom;
+    var i;
+    var labelKeys;
+    var labelKey;
 
     Object.keys(simple_map).forEach(function (k) {
         if (container.hasOwnProperty(k) && container[k].length > 0) {
@@ -1507,15 +1507,28 @@ function buildVmPayload(opts, container, callback) {
         });
     }
     if (container.Labels) {
-        Object.keys(container.Labels).forEach(function _labelToTag(key) {
-            if (key === PACKAGE_SELECTION_LABEL) {
+        labelKeys = Object.keys(container.Labels);
+        for (i = 0; i < labelKeys.length; i++) {
+            labelKey = labelKeys[i];
+            if (labelKey === PACKAGE_SELECTION_LABEL) {
                 // PACKAGE_SELECTION_LABEL is a special label. It gets consumed
                 // here and doesn't end up in the user's list. Instead it's used
                 // in getPackage().
-                return;
+                continue;
             }
-            payload.tags[common.LABELTAG_PREFIX + key] = container.Labels[key];
-        });
+            if (triton_tags.isTritonTag(labelKey)) {
+                try {
+                    payload.tags[labelKey] = triton_tags.validateTritonTagStr(
+                        labelKey, container.Labels[labelKey]);
+                } catch (err) {
+                    callback(err);
+                    return;
+                }
+            } else {
+                payload.tags[common.LABELTAG_PREFIX + labelKey]
+                    = container.Labels[labelKey];
+            }
+        }
     }
 
     if (container.Cmd) {
--- a/lib/backends/sdc/containers.js
+++ b/lib/backends/sdc/containers.js
@@ -1472,9 +1472,6 @@ function buildVmPayload(opts, container, callback) {
     };
     var restartPolicy;
     var volumesFrom;
-    var i;
-    var labelKeys;
-    var labelKey;
 
     Object.keys(simple_map).forEach(function (k) {
         if (container.hasOwnProperty(k) && container[k].length > 0) {
@@ -1500,36 +1497,6 @@ function buildVmPayload(opts, container, callback) {
     payload.tags = {
         'sdc_docker': true
     };
-    // Add docker labels as prefixed tags on the vm.
-    if (imgConfig.Labels) {
-        Object.keys(imgConfig.Labels).forEach(function (key) {
-            payload.tags[common.LABELTAG_PREFIX + key] = imgConfig.Labels[key];
-        });
-    }
-    if (container.Labels) {
-        labelKeys = Object.keys(container.Labels);
-        for (i = 0; i < labelKeys.length; i++) {
-            labelKey = labelKeys[i];
-            if (labelKey === PACKAGE_SELECTION_LABEL) {
-                // PACKAGE_SELECTION_LABEL is a special label. It gets consumed
-                // here and doesn't end up in the user's list. Instead it's used
-                // in getPackage().
-                continue;
-            }
-            if (triton_tags.isTritonTag(labelKey)) {
-                try {
-                    payload.tags[labelKey] = triton_tags.validateTritonTagStr(
-                        labelKey, container.Labels[labelKey]);
-                } catch (err) {
-                    callback(err);
-                    return;
-                }
-            } else {
-                payload.tags[common.LABELTAG_PREFIX + labelKey]
-                    = container.Labels[labelKey];
-            }
-        }
-    }
 
     if (container.Cmd) {
         payload.internal_metadata['docker:cmd'] = JSON.stringify(container.Cmd);
@@ -1912,6 +1879,74 @@ function buildVmPayload(opts, container, callback) {
             });
 
             cb();
+        },
+
+        /**
+         * Add VM payload tags for image and container labels.
+         *
+         * - package-selection label is special (handled separately below)
+         * - "triton." tags are parsed/validated and added without a prefix
+         * - all others are added with "docker:label:" prefix
+         */
+        function vmTagsFromLabels(_, cb) {
+            assert.object(payload.tags, 'payload.tags');
+
+            var labels = [];
+            if (imgConfig.Labels) {
+                Object.keys(imgConfig.Labels).forEach(function (k) {
+                    labels.push([k, imgConfig.Labels[k]]);
+                });
+            }
+            if (container.Labels) {
+                Object.keys(container.Labels).forEach(function (k) {
+                    labels.push([k, container.Labels[k]]);
+                });
+            }
+
+            var labelErrs = [];
+            vasync.forEachPipeline({
+                inputs: labels,
+                func: function handleLabel(label, next) {
+                    var key = label[0];
+                    var val = label[1];
+
+                    if (key === PACKAGE_SELECTION_LABEL) {
+                        /*
+                         * PACKAGE_SELECTION_LABEL is a special label. It gets
+                         * consumed here and doesn't end up in the user's list.
+                         * It is handled below in `selectPackage()`.
+                         */
+                        next();
+                    } else if (triton_tags.isTritonTag(key)) {
+                        triton_tags.parseTritonTagStr(key, val,
+                                function (err, parsed) {
+                            if (err) {
+                                labelErrs.push(err);
+                            } else {
+                                payload.tags[key] = parsed;
+                            }
+                            next();
+                        });
+                    } else {
+                        payload.tags[common.LABELTAG_PREFIX + key] = val;
+                        next();
+                    }
+                }
+            }, function (err) {
+                if (err) {
+                    cb(err);
+                } else if (labelErrs.length > 0) {
+                    cb(new errors.ValidationError(util.format(
+                        'invalid label%s: %s',
+                        (labelErrs.length === 1 ? '' : 's'),
+                        labelErrs.map(function (e) { return e.message; })
+                            .join('; '))));
+                } else {
+                    log.trace({labels: labels, tags: payload.tags},
+                        'buildVmPayload: vmTagsFromLabel');
+                    cb();
+                }
+            });
         },
 
         function selectPackage(_, cb) {
--- a/lib/backends/sdc/utils.js
+++ b/lib/backends/sdc/utils.js
@@ -5,32 +5,71 @@
  */
 
 /*
- * Copyright (c) 2015, Joyent, Inc.
+ * Copyright 2016, Joyent, Inc.
  */
 
 var assert = require('assert-plus');
+var format = require('util').format;
+var mod_fwrule = require('fwrule');
 var net = require('net');
 var restify = require('restify');
 
 var common = require('../../common');
 var constants = require('../../constants');
 var Link = require('../../models/link');
+var triton_tags = require('../../triton-tags');
 
-var mod_fwrule = require('fwrule');
 
 
 function vmUuidToShortDockerId(uuid) {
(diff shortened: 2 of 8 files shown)


lib/backends/sdc/containers.js
```diff
@@ -1475,6 +1472,9 @@ function buildVmPayload(opts, container, callback) {
     };
     var restartPolicy;
     var volumesFrom;
+    var i;
+    var labelKeys;
+    var labelKey;
```
joshwilsdon: These aren't in the correct place alphabetically.

lib/backends/sdc/containers.js
```diff
@@ -1507,15 +1507,28 @@ function buildVmPayload(opts, container, callback) {
...
+                continue;
             }
-            payload.tags[common.LABELTAG_PREFIX + key] = container.Labels[key];
-        });
+            if (triton_tags.isTritonTag(labelKey)) {
+                try {
+                    payload.tags[labelKey] = triton_tags.validateTritonTagStr(
+                        labelKey, container.Labels[labelKey]);
+                } catch (err) {
+                    callback(err);
```
joshwilsdon: Would it be good to turn this into a more specific error so we don't end up with a generic one to the user?

lib/backends/sdc/utils.js
```diff
@@ -5,32 +5,71 @@
...
+ * Note: Docker label values must be strings otherwise the docker 1.10 client
+ * will blow up. See https://github.com/docker/docker/issues/20881 and
+ * DOCKER-721. Currently VM.tags are limited to string, boolean or number,
+ * so we can simply `.toString()`.
+ */
 function dockerLabelsFromVmTags(tags) {
     var labels = {};
-    var labelPrefix = common.LABELTAG_PREFIX;
+    const LABELTAG_PREFIX = common.LABELTAG_PREFIX;
+    const expectedTypes = {string: true, boolean: true, number: true};
```
joshwilsdon: should these be alphabetical?

lib/backends/sdc/containers.js
trentm: Ah didn't know we were doing alpha. Will update.

lib/backends/sdc/containers.js
trentm: Yes definitely. Thanks. I'd forgotten to do that.

FWIW, I left triton-tags.js generic (i.e. not using the local errors.js) so that it could eventually be pulled out to a separate module (node-triton-tags.js or whatever) to have the validation shared between here and in VMAPI.

lib/backends/sdc/utils.js
trentm: will do

joshwilsdon: ooc: why object for expectedTypes instead of array?

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+ * For now, using the more limited labels allowed by RFC1123. RFC2181 supercedes
+ * 1123, but the broader range of characters can sometimes cause problems with
+ * other systems (e.g. see the underscore in RFC5321).
+ */
+var DNS_NAME_RE = /^[a-z0-9][a-z0-9\-]{0,62}(?:\.[a-z0-9][a-z0-9\-]{0,62})*$/i;
+
+const TRITON_TAG_PREFIX = 'triton.';
+
+var typeFromKey = {
+    'triton.cns.services': 'string',
```
joshwilsdon: should these be:

```
TRITON_TAG_PREFIX + '.cns.services': 'string',
```

?

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+    'triton.cns.reverse_ptr': 'string'
+};
+
+/*
+ * Validator functions take a `val` (already checked to be of the appropriate
+ * type per `typeFromKey`) and should throw an `Error` if invalid. The don't
+ * return a value.
+ */
+var validatorFromKey = {
+    'triton.cns.services': function validateTritonCnsServices(val) {
```
joshwilsdon: TRITON_TAG_PREFIX + '.cns.services'?

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+        for (var i = 0; i < fqdns.length; i++) {
+            var fqdn = fqdns[i];
+            if (fqdn.length > 255 || !fqdn.match(DNS_NAME_RE)) {
+                unsafes.push(fqdn);
+            }
+        }
+
+        if (unsafes.length > 0) {
+            throw new Error(format(
+                'invalid "triton.cns.services" tag: "%s" %s not DNS safe',
```
joshwilsdon: TRITON_TAG_PREFIX?

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+
+var typeFromKey = {
+    'triton.cns.services': 'string',
+    'triton.cns.disable': 'boolean',
+    'triton.cns.reverse_ptr': 'string'
+};
+
+/*
+ * Validator functions take a `val` (already checked to be of the appropriate
+ * type per `typeFromKey`) and should throw an `Error` if invalid. The don't
```
joshwilsdon: *They don't

lib/backends/sdc/containers.js
```diff
@@ -1507,15 +1507,28 @@ function buildVmPayload(opts, container, callback) {
...
                 // PACKAGE_SELECTION_LABEL is a special label. It gets consumed
                 // here and doesn't end up in the user's list. Instead it's used
                 // in getPackage().
-                return;
+                continue;
             }
-            payload.tags[common.LABELTAG_PREFIX + key] = container.Labels[key];
-        });
+            if (triton_tags.isTritonTag(labelKey)) {
+                try {
```
joshwilsdon: By having these validateTritonTag* functions throw and catching in this `try {`, we lose the ability to distinguish between programmer errors and runtime errors. For example we've validated that we have a triton tag here, but later we'll do:

```
assert.func(validator, 'validator for tag ' + key);
```

or:

```
if (type === undefined) {
    throw new Error('Unrecognized special triton tag "' + key + '"');
}
```

which seem like they would only happen on programmer error. But they'll be caught in this `try {` rather than throwing an uncaught exception and generating a core. Is this fine? It might be, just wanted to point it out.

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+        }
+    }
+};
+
+
+// --- exports
+
+/*
+ * Return true if the given key uses the Triton tag prefix.
+ * Note that it still might not be one of the specific defined tag
```
joshwilsdon: *tag**s**

lib/backends/sdc/utils.js
trentm: easier lookup that arr.indexOf, basically. I don't have a string pref.
If JS had Python's `in` operator, I'd use an array for this kind of thing much more frequently.

lib/triton-tags.js
trentm: I suppose, yes. Though I'd be a little sad. I expect more grepping for "triton.cns.services" than for "cns.services".

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+    assert.func(validator, 'validator for tag ' + key);
+    validator(val);
+
+    return val;
+}
+
+/*
+ * Convert and validate the given Triton tag string value.
+ *
+ * @returns {String} On success, the validated value is returned.
```
joshwilsdon: This seems incorrect in that the return will not always be a String. It will be one of String/Boolean/Number, correct?

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+
+    var validator = validatorFromKey[key];
+    assert.func(validator, 'validator for tag ' + key);
+    validator(val);
+
+    return val;
+}
+
+/*
+ * Convert and validate the given Triton tag string value.
```
joshwilsdon: Might be good to be a bit more specific about the conversion intended here.

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+        } else if (str === 'false') {
+            val = false;
+        } else {
+            throw new Error(format(
+                'Triton tag "%s" value must be "true" or "false": %j',
+                key, str));
+        }
+        break;
+    case 'number':
+        val = Number(str);
```
joshwilsdon: Do we care about an empty string here? Number('') will be 0.

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+    // Validate.
+    var validator = validatorFromKey[key];
+    assert.func(validator, 'validator for tag ' + key);
+    validator(val);
+
+    return val;
+}
+
+
+module.exports = {
```
joshwilsdon: Should this also export TRITON_TAG_PREFIX in case callers (like cli-triton-tags.test.js) want to use the prefix, so they don't also need to hardcode this value?

lib/backends/sdc/containers.js
trentm: Exploring that, which do you like best:

```
// Returns value, or throws Error.
try {
    payload.tags[labelKey] = triton_tags.validateTritonTagStr(
        labelKey, container.Labels[labelKey]);
} catch (err) {
    callback(new errors.ValidationError(err, err.message));
    return;
}

// Returns one value: the validated value or an Error instance.
var errOrVal = triton_tags.validateTritonTagStr(
    labelKey, container.Labels[labelKey]);
if (errOrVal instanceof Error) {
    callback(new errors.ValidationError(errOrVal,
        errOrVal.message));
    return;
} else {
    payload.tags[labelKey] = errOrVal;
}

// Returns a tuple `[err, val]`.
var rv = triton_tags.validateTritonTagStr(   // [err, val]
    labelKey, container.Labels[labelKey]);
if (rv[0]) {
    callback(new errors.ValidationError(rv[0], rv[0].message));
    return;
} else {
    payload.tags[labelKey] = rv[1];
}
```

?

lib/triton-tags.js
trentm: You are correct. Updated to:

```
 * @returns {String|Boolean|Number} On success, the validated value is returned.
```

lib/triton-tags.js
trentm: Perhaps call it `parseTritonTagStr` (and leave the other one as `validateTritonTag` -- that one intended to be used by VMAPI which gets tags with the proper type)?

lib/triton-tags.js
trentm: I also forgot to give isNaN an arg there. Corrected:
```
        if (isNaN(val)) {
```

lib/triton-tags.js
```diff
@@ -0,0 +1,185 @@
...
+ * return a value.
+ */
+var validatorFromKey = {
+    'triton.cns.services': function validateTritonCnsServices(val) {
+        assert.string(val, 'val');
+
+        var unsafes = [];
+        var fqdns = val.split(',');
+        for (var i = 0; i < fqdns.length; i++) {
+            var fqdn = fqdns[i];
```
richardkiene: This logic looks like it came from VMAPI and isn't quite correct. I have a change I'll be merging in (thanks to @arekinath) to VMAPI shortly and we'll want that same logic here if it must be copied. The new change allows for SRV records to be used with CNS (e.g. triton.cns.services=jabber:5223).

Should we think about a way to better re-use code?

lib/triton-tags.js
trentm: Thanks. I'll get it to disallow empty string.

lib/backends/sdc/containers.js
joshwilsdon: I'd propose 2 more options, first the one I'd probably use since I believe it's the idiomatic node way to do this:

```
// Calls callback(err, value)
triton_tags.validateAndNormalizeTritonTagStr(labelKey, container.Labels[labelKey], function _validateCb(err, value) {
    if (err) {
        callback(err);
        return;
    }

    payload.tags[labelKey] = value;
});
```

or if the reason for it being invalid isn't important (like the current code):

```
// separate validation and normalization
if (triton_tags.validTritonTagStr(labelKey, container.Labels[labelKey])) {
    payload.tags[labelKey] = triton_tags.normalizeTritonTagStr(labelKey, container.Labels[labelKey]));
} else {
    callback(new errors.ValidationError('invalid ' + labelKey));
    return;
}
```

test/integration/cli-triton-tags.test.js
```diff
@@ -0,0 +1,159 @@
...
+        cli.run(t, {
+            args: args,
+            /* JSSTYLED */
+            expectedErr: 'Error response from daemon: Unrecognized special triton tag "triton.bogus"'
+        });
+    });
+});
+
+
+test('teardown', cli.rmAllCreated);
```
joshwilsdon: We seem to have tested boolean and string values, should we also add a test for a triton.cns tag that has a numeric value so we test the:

```
case 'number':
```

codepath? With both a valid and invalid number and maybe a float and an empty string?

trentm: @joshwilsdon There aren't any current `triton.` tags that take a number.

lib/triton-tags.js
trentm: Is there a case where it is needed? I actually intentionally didn't export it: users of the module should use `isTritonTag(key)` instead.

lib/triton-tags.js
trentm: I'm inclined to not do this step, because JS doesn't allow an expression for the key part of an object literal:

```
var typeFromKey = {
    TRITON_TAG_PREFIX + 'cns.services': 'string',   // <--- syntax error
...
```

I wonder if an acceptable alternative (if really wanted) is a top-level assert that all the keys in the
`typeFromKey` and `validatorFromKey` start with the prefix?

test/lib/vm.js
```diff
@@ -51,6 +51,104 @@ function getVm(t, opts, callback) {
...
+            var keys = Object.keys(opts.tags);
+            for (var i = 0; i < keys.length; i++) {
+                var k = keys[i];
+                if (obj.tags[k] !== opts.tags[k]) {
+                    incomplete = true;
+                    break;
+                }
+            }
+
+            if (!incomplete) {
```
joshwilsdon: Since this doesn't check the job (which might have failed) this appears it could continue forever. Perhaps either add a check on the job 'execution' field or at least a timeout value such that if this runs say some hundreds of loops we'll give up?

lib/triton-tags.js
joshwilsdon: That's fine. I'd suggest taking out the TRITON_TAG_PREFIX then and just using 'triton.' everywhere. So that someone doesn't think they can actually change this meaningfully in one place. Or add a comment saying that this is only a convenience thing and in order to change one would still need to replace all instances of `triton.` everywhere.

lib/triton-tags.js
joshwilsdon: yeah, `parseTritonTagStr` I think fits better.

test/integration/cli-triton-tags.test.js
joshwilsdon: Perhaps we shouldn't include the code to support that yet then? Since it can't be tested. It seems better for validateTritonTagStr to just not support numbers yet than to have code there that hasn't and can't be tested.

lib/triton-tags.js
joshwilsdon: The only reason would be if you want to be able to change it. But as I said above, that's the only good reason I see for having a constant at all. Since we're not going to make it possible to change, there's no good reason to export it. Everyone can just hardcode 'triton.' and be happy.

lib/triton-tags.js
```diff
@@ -0,0 +1,229 @@
...
+        cb(new Error(errmsg));
+    } else {
+        cb(null, val);
+    }
+}
+
+
+module.exports = {
+    isTritonTag: isTritonTag,
+    validateTritonTag: validateTritonTag,
```
joshwilsdon: Where is validateTritonTag() actually used? I must be missing something. Why is this needed?

trentm: It isn't yet. It is the intended function to be used by VMAPI later today when I break this
out to a separate node-triton-tags module and have docker and vmapi share that.
However, I'd like to get this in sdc-docker first.

joshwilsdon:
Aside from perhaps some tests for the validateTritonTag() function (or removing from the API until such time as it's needed), I'd say this lgtm.

trentm:
Squashed and merged to master in commit 5e655d4

PR #71, opened 2016-03-15
DOCKER-735: queryParser handlers missing for some endpoints that require the query string to be parsed

misterdjules:
The following endpoints:
- containerStop
- containerKill
- containerRestart

were missing a `queryParser` handler to parse the query string. As a result, none of their query string parameters were used. This problem was the result of a previous change that moved the `queryParser` from the `before` handlers to each specific endpoint that was using it.

This PR adds the `queryParser` handler to these three endpoints, and adds some basic tests that were missing, and that would have helped us find that the `queryParser` handler was missing in the first place.

In addition to these changes, this PR brings the following other notable changes:
1. It adds a check in error classes to make sure that their constructor is called with new, preventing hard to find bugs.
2. It adds some validation for timeout values in the `containerStop` and `containerRestart` endpoints, and for signal values in the `containerKill` endpoint.
3. It adds a new `test/lib/container.js` module that contains the `checkContainerStatus` function. This function is used to check if a container transitions to a given status and allows for a certain number of retries. This was needed in the `api-start-kill.js` integration test to make it non-flakey, since sometimes a killed container would not transition to the `stopped` status before we'd check its status.

This PR needs to land _after https://github.com/joyent/sdc-vmapi/pull/22, since otherwise it would pass numeric signal such as `'9'` as strings, which VMAPI doesn't accept before https://github.com/joyent/sdc-vmapi/pull/22.

### DIFF
--- a/lib/validate.js
+++ b/lib/validate.js
@@ -431,7 +431,147 @@ function validateArchiveWriteStream(req, res, next) {
     next();
 }
 
+/*
+ * Validates that the query string parameter "paramName" is a valid signal,
+ * and calls "callback" with an error parameter and the validated value. For
+ * consistency, the validated value is always a string, even if the signal is
+ * represented by a number.
+ */
+function validateSignal(req, paramName, callback) {
+    assert.object(req, 'req');
+    assert.object(req.query, 'req.query');
+    assert.string(paramName, 'paramName');
+    assert.func(callback, 'callback');
 
+    var signalString;
+    var validSignal = false;
+    var err;
+
+    // XXX This includes all signals from process.binding('constants') plus
+    // the special '0' signal. Anything not on this list is not supported by
+    // the backend, but it's possible we'll want to remove some from this list
+    // too as some of these probably don't make sense to send.
+    var validSignals = [
+        'SIGABRT', 'SIGALRM', 'SIGBUS', 'SIGCHLD', 'SIGCONT', 'SIGFPE',
+        'SIGHUP', 'SIGILL', 'SIGINT', 'SIGIO', 'SIGIOT', 'SIGKILL', 'SIGLOST',
+        'SIGPIPE', 'SIGPOLL', 'SIGPROF', 'SIGPWR', 'SIGQUIT', 'SIGSEGV',
+        'SIGSTOP', 'SIGSYS', 'SIGTERM', 'SIGTRAP', 'SIGTSTP', 'SIGTTIN',
+        'SIGTTOU', 'SIGURG', 'SIGUSR1', 'SIGUSR2', 'SIGVTALRM', 'SIGWINCH',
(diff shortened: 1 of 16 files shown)


lib/validate.js
```diff
@@ -431,7 +431,147 @@ function validateArchiveWriteStream(req, res, next) {
...
+        'SIGABRT', 'SIGALRM', 'SIGBUS', 'SIGCHLD', 'SIGCONT', 'SIGFPE',
+        'SIGHUP', 'SIGILL', 'SIGINT', 'SIGIO', 'SIGIOT', 'SIGKILL', 'SIGLOST',
+        'SIGPIPE', 'SIGPOLL', 'SIGPROF', 'SIGPWR', 'SIGQUIT', 'SIGSEGV',
+        'SIGSTOP', 'SIGSYS', 'SIGTERM', 'SIGTRAP', 'SIGTSTP', 'SIGTTIN',
+        'SIGTTOU', 'SIGURG', 'SIGUSR1', 'SIGUSR2', 'SIGVTALRM', 'SIGWINCH',
+        'SIGXCPU', 'SIGXFSZ', 'ABRT', 'ALRM', 'BUS', 'CHLD', 'CONT', 'FPE',
+        'HUP', 'ILL', 'INT', 'IO', 'IOT', 'KILL', 'LOST', 'PIPE', 'POLL',
+        'PROF', 'PWR', 'QUIT', 'SEGV', 'STOP', 'SYS', 'TERM', 'TRAP', 'TSTP',
+        'TTIN', 'TTOU', 'URG', 'USR1', 'USR2', 'VTALRM', 'WINCH', 'XCPU',
+        'XFSZ', '0', '1', '2', '3', '4', '5', '6', '6', '8', '9', '10', '11',
```
joshwilsdon: @trentm pointed out in sdc-vmapi#22 that there are duplicates. Looks like the same problem exists here.

misterdjules:
Updated this PR with [a second commit](https://github.com/joyent/sdc-docker/pull/71/commits/4a3504ad9f6dbb320c3e205a3d175f65fb69386c) that:

* Uses only symbolic signal names as strings when sending requests to VMAPI even when numeric signals are received as input, so this PR no longer depends on joyent/sdc-vmapi#22.

* Adds a lot of integration tests for the kill endpoint. These tests use a custom Docker image that only echoes to stdout any signal that can be handled by a custom signal handler. This image was pushed to Docker Hub.

* Adds a new signal module in lib/signals that comes with unit tests.

* Adds a few functions to `test/lib/cli.js` like `cli.exec`, `cli.logs` and `cli.kill`. `cli.exec` replaces the previous `cli.exec` which was renamed to `cli.execInTestZone`.

misterdjules:
Closing as I won't be working on that and I don't think there's any interest in getting that merged.

PR #75, opened 2016-04-04
DOCKER-523: Implement docker commit

twhiteman:


lib/backends/sdc/build.js
```diff
@@ -891,8 +894,266 @@ function runBuildCommand(opts, callback) {
...
+/**
+ * Commit (create) an image from a given container.
+ *
+ * @param {Object} opts
+ * @param {String} opts.author - Author name for the image
+ * @param {Array} opts.changes - Docker image changes to make.
+ * @param {String} opts.comment - Comment notes for the image.
+ * @param {Object} opts.config - Image config object
+ * @param {Boolean} opts.pause - common.boolFromQueryParam(req.query.pause),
+ * @param {Object} opts.req - Request instance.
```
misterdjules: Is `req` in `opts.req` or just in the first `req` parameter?

lib/backends/sdc/build.js
```diff
@@ -891,8 +894,266 @@ function runBuildCommand(opts, callback) {
...
+    assert.object(vm, 'vm');  // Should have been added by the backend.
+
+    var baseImg;
+    var baseImageId = vm.internal_metadata['docker:imageid'];
+    var cleanupTimeoutId = -1;
+    var commitFinishedSuccessfully = false;
+    var finalImageId = null;  // Final docker image id for resulting image.
+    var log = req.log;
+    var previousErr;
+    var rat;
```
misterdjules: What does `rat` mean in this context?

misterdjules: Ah, "repo and tag" :)

lib/backends/sdc/build.js
```diff
@@ -891,8 +894,266 @@ function runBuildCommand(opts, callback) {
...
+    function getScratchImage(_, cb) {
+        req.app.backend.getScratchImage(req,
+        function _getScratchImageCb(err, sImg) {
+            if (err) {
+                log.error(err, 'getScratchImage failure');
+                cb(err);
+                return;
+            }
+            scratchImage = sImg;
+            //log.debug('scratchImage: ', scratchImage);
```
misterdjules: Is there any problem in keeping that debug log instead of commenting it out?

lib/backends/sdc/build.js
```diff
@@ -891,8 +894,266 @@ function runBuildCommand(opts, callback) {
...
+    }
+
+
+    function runCnapiDockerCommit(_, cb) {
+        var commitError;
+        var commitEventStream;
+        var host = socketOpts.host;
+        var port = socketOpts.port;
+
+        var commitSocket = net.createConnection({ host: host, port: port });
```
misterdjules: We may want to register an `'error'` event listener, and maybe a `'timeout'` event listener too.

lib/backends/sdc/build.js
```diff
@@ -891,8 +894,266 @@ function runBuildCommand(opts, callback) {
...
+        });
+        commitEventStream.on('close', function _eventStreamCloseCb() {
+            log.debug('commit: commitEventStream.close - closing socket');
+        });
+        commitSocket.on('end', function _commitSocketEndCb() {
+            log.debug('commit: commitSocket.end', commitError);
+            cb(commitError);
+        });
+
+        log.debug('commit: creating second connection to cnapi');
```
misterdjules: Shouldn't that log message be outputted close to the call to `net.createConnection` at the beginning of the function so that it's clearer what it refers to?

lib/backends/sdc/build.js
twhiteman: Just as the first argument - will update that comment. Note: opts in this case should be what was supplied via the docker client (i.e. query params).

lib/backends/sdc/build.js
```diff
@@ -891,8 +894,266 @@ function runBuildCommand(opts, callback) {
...
+
+
+    function recordError(err, logMessage) {
+        log.error(err, logMessage);
+        if (!previousErr) {
+            previousErr = err;
+            // If in 5 minutes the cn-agent task hasn't gotten back to us, then
+            // something is off the rails and we'll abort/cleanup from here by
+            // firing the callback, thus closing the client.
+            cleanupTimeoutId = setTimeout(commitCleanup, 5 * 60 * 1000);
```
misterdjules: We might be able to not have to setup a custom timer by adding a `'timeout'` event listener on the socket as mentioned above. If the default timeout set by node for sockets is not suitable, we can set a custom timeout value with `setTimeout(timeoutValue)` on the socket.

lib/backends/sdc/build.js
twhiteman: Yes, same name is used elsewhere in images - blame Trent ;)

lib/backends/sdc/build.js
twhiteman: I found it spammy - I will drop that line (rather than comment it out). It doesn't give much additional information for docker commit (as the only place it's used is a reference check), and the contents are static (already defined in the *etc/scratch_image.json* file).

lib/backends/sdc/build.js
twhiteman: Good suggestion - admittedly this (socket events) is not my strong point in node - do you have an example of these? If not I'll dig.

lib/backends/sdc/images.js
```diff
@@ -643,6 +646,86 @@ function imgFromName(opts, callback) {
 
 
 /**
+ * Return the docker `Image` model for the given the given IMGAPI uuid.
```
misterdjules: Nit: repetition of `the given`.

lib/endpoints/commit.js
```diff
@@ -5,19 +5,68 @@
...
+    };
+
+    log.debug('commit: req.query: ', req.query);
+
+    /*
+     * Node's default HTTP timeout is two minutes, and the client context
+     * request can take longer than that to complete.  Set this connection's
+     * timeout to an hour to avoid an abrupt close after two minutes.
+     */
+    req.connection.setTimeout(60 * 60 * 1000);
```
misterdjules: Maybe a timeout of an hour is a bit much, and might actually be used to DOS the service by creating a lot of requests that never time out. I know the same timeout is used for other requests, and a shorter timeout would not necessarily solve the issue entirely, but could we use something shorter?

lib/endpoints/commit.js
```diff
@@ -5,19 +5,68 @@
...
+
+    req.backend.commitImage(req, commitOpts, function (err, imageId) {
+        if (err) {
+            res.send(err);
+        } else {
+            log.debug('docker commit finished successfully');
+            // Note: Docker returns sha256: {"Id":"sha256:93c3e0ca32...1453a"}
+            res.write(JSON.stringify({'Id': imageId}) + '\n');
```
misterdjules: What is the HTTP status code of that response, do we want it to be something specific like `201` for "a new resource has been created", or does Docker expects a 200?

lib/endpoints/commit.js
```diff
@@ -5,19 +5,68 @@
...
+        }
+
+        res.end();
+        next(true);
```
misterdjules: What does passing `true` to next mean?

lib/backends/sdc/build.js
twhiteman: Yes, moved.

test/integration/cli-commit.test.js
```diff
@@ -0,0 +1,117 @@
...
+    tt.test('delete ' + IMAGE_NAME + ' container', function (t) {
+        cli.rm(t, {args: containerName}, t.end.bind(t)); // Err handled in rm
+    });
+
+    // Inspect committed image metadata.
+    tt.test('inspect committed image', function (t) {
+        cli.inspect(t, {
+            id: commitImageTag
+        }, function (err, img) {
+            if (img) {
```
misterdjules: Should we add a `t.ifErr(err)` and a `t.ok(img)` to catch the case when there's an error and/or there's no image?

test/integration/cli-commit.test.js
```diff
@@ -0,0 +1,117 @@
...
+    });
+
+    // Run the committed image and verify the 'newfile.txt' contents.
+    tt.test('verify created image', function (t) {
+        var runArgs = format('--rm --name %s %s sh -c "cat /newfile.txt"',
+            common.makeContainerName(CONTAINER_PREFIX + 'verify_'),
+            commitImageTag);
+        cli.run(t, {args: runArgs}, function (err, result) {
+            // err is already tested in cli.run() call
+            if (!err) {
```
misterdjules: Should we add a `t.ifErr(err)` to catch the case when there's an error?

test/integration/cli-labels.test.js
```diff
@@ -13,12 +13,12 @@
  */
 
 var format = require('util').format;
-var libuuid = require('libuuid');
 var test = require('tape');
 var vasync = require('vasync');
 
 var cli = require('../lib/cli');
 var vm = require('../lib/vm');
+var common = require('../lib/common');
```
misterdjules: Nit: move common above `vm` to keep requires sorted alphabeticaly.

lib/endpoints/commit.js
twhiteman: Yes, I will make the commit timeout shorter (as it's not expected to do as much work as say, docker build) - probably 5-10 minutes will suffice.

lib/endpoints/commit.js
twhiteman: Good spotting - should be 201.

test/integration/cli-commit.test.js
```diff
@@ -0,0 +1,117 @@
...
+var h = require('./helpers');
+
+var format = require('util').format;
+
+var STATE = {
+    log: require('../lib/log')
+};
+
+var CONTAINER_PREFIX = 'sdcdockertest_commit_';
+var DOCKER_ALICE; // Regular JSON restify client.
```
misterdjules: `DOCKER_ALICE` doesn't seem to be used.

lib/endpoints/commit.js
twhiteman: I'll change that to *next()*. I had thought it meant that we are done with the request and there are no more handlers to run for this restify route, but reading the restify docs and code I find it confusing.

test/integration/cli-commit.test.js
twhiteman: `t.ifErr(err)` check is already in the cli.inspect() code, as well as some checking for valid JSON response.

test/integration/cli-commit.test.js
twhiteman: Same - as my comment `// err is already tested in cli.run() call` :P

test/integration/cli-commit.test.js
twhiteman: Yub, not used any more - gone.

lib/backends/sdc/build.js
misterdjules: Trent blamed! :D

lib/backends/sdc/build.js
misterdjules: It follows the usual pattern of:

```
commitSocket.on('error', function onError(err) {});
```
and:
````
commitSocket.on('timeout', function onTimeout(){});
````

Note that in the case of the 'timeout' event the socket itself is not closed or modified in any way, so if the desired behavior is to close it, then it'll need to be done manually.

PR #77, opened 2016-04-12
DOCKER-787 catch error when connecting for attach/exec

joshwilsdon:
When a connection to the agent fails with ECONNREFUSED, the restify domain's handler catches the error and we end up not calling /wait from the client. This means that:

```
docker run --rm node
```

sometimes works and sometimes bails because the container is still running (when we haven't run wait to see it stop). This fix makes it so that when a connection is refused we get the normal 400 response instead and the client calls wait.

### DIFF
--- a/lib/backends/sdc/containers.js
+++ b/lib/backends/sdc/containers.js
@@ -3535,6 +3535,9 @@ function _runExec(opts, callback) {
     assert.object(opts.socketData, 'opts.socketData');
     assert.object(opts.socket, 'opts.socket');
 
+    // Make sure our callbacks get called only once
+    var cb = once(callback);
+
     var socketData = opts.socketData;
     var host = socketData.host;
     var port = socketData.port;
@@ -3572,18 +3575,17 @@ function _runExec(opts, callback) {
         serverSocket.end();
         cb(error, socketData);
     }
-    // Make sure our callbacks get called only once
+
     var endSocket = once(_endSocket);
-    var cb = once(callback);
 
     serverSocket.on('connect', setupListeners);
 
     // error can happen before connect too (eg. ECONNREFUSED)
-    serverSocket.on('error', function (error) {
+    serverSocket.on('error', function _onServerSocketError(error) {
         opts.log.error('serverSocket for %s threw an error %s',
             cmdString, error.toString());
 
-        endSocket(error);
+        cb(error);
     });
 
     function setupListeners() {
@@ -3655,7 +3657,6 @@ function _runAttach(opts, callback) {
             opts.log.debug('attach for %s threw an error %',
                 cmdString, error.toString());
 
-            serverSocket.end();
             cb(error);
         });
 
--- a/lib/backends/sdc/containers.js
+++ b/lib/backends/sdc/containers.js
@@ -3578,6 +3578,14 @@ function _runExec(opts, callback) {
 
     serverSocket.on('connect', setupListeners);
 
+    // error can happen before connect too (eg. ECONNREFUSED)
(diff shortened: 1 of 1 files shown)


lib/backends/sdc/containers.js
```diff
@@ -3578,6 +3578,14 @@ function _runExec(opts, callback) {
 
     serverSocket.on('connect', setupListeners);
 
+    // error can happen before connect too (eg. ECONNREFUSED)
+    serverSocket.on('error', function (error) {
```
misterdjules: While we're at it, maybe we should name that function.

lib/backends/sdc/containers.js
```diff
@@ -3578,6 +3578,14 @@ function _runExec(opts, callback) {
...
+        opts.log.error('serverSocket for %s threw an error %s',
+            cmdString, error.toString());
+
+        endSocket(error);
```
misterdjules: Why do we use `endSocket` here and not in `_runAttach`? It seems that `endSocket` would display a message such as `'client socket threw an error`, but here we're dealing with a socket that is connected to the CN. So is `serverSocket.end()` here more appropriate?

lib/backends/sdc/containers.js
```diff
@@ -3647,6 +3651,14 @@ function _runAttach(opts, callback) {
     if (!serverSocket) {
         serverSocket = net.createConnection({ host: host, port: port });
         serverSocket.on('connect', setupListeners);
+        serverSocket.on('error', function (error) {
+            opts.log.debug('attach for %s threw an error %',
+                cmdString, error.toString());
+
+            serverSocket.end();
```
misterdjules: I don't think `serverSocket` is needed here, was it meant to close the other side of the connection (in that case `clientSocket` if I understand correctly)?

joshwilsdon: You mean serverSocket.end() is not needed? If so, I'll remove from the other call as well.

lib/backends/sdc/containers.js
joshwilsdon: The reason is just that they're in different scopes. _runAttach will be run potentially multiple times (by multiple attaching clients) and the endSocket function there is within the scope of the setupListeners() function. It also has a comment:

```
        // When using multi attach the _endSocket callback will be GC'd
        // if defined outside of this scope
```

which made me think changing the scope wasn't a great idea. You're probably right that they should do the same thing though. Does changing this to not call endSocket and instead just doing:

```
    // error can happen before connect too (eg. ECONNREFUSED)
    serverSocket.on('error', function _onServerSocketError(error) {
        opts.log.error('serverSocket for %s threw an error %s',
            cmdString, error.toString());

        serverSocket.end();
        cb(error);
    });
```

make more sense?

lib/backends/sdc/containers.js
misterdjules: Sorry yes, `serverSocket.end()` is not needed.

lib/backends/sdc/containers.js
misterdjules: > Does changing this to not call endSocket and instead just doing:

    // error can happen before connect too (eg. ECONNREFUSED)
    serverSocket.on('error', function _onServerSocketError(error) {
        opts.log.error('serverSocket for %s threw an error %s',
            cmdString, error.toString());

        serverSocket.end();
        cb(error);
    });
make more sense?

Yes, and I don't think the call to `serverSocket.end()` is needed here too.

misterdjules:
@joshwilsdon LGTM :)

PR #78, opened 2016-04-13
DOCKER-793: First pass for NFS shared volumes

misterdjules:
As indicated by the title of this PR, this is a first pass so that we can enable experimental support for NFS shared volumes at least in COAL, and hopefully in nightly. There's a lot of naive assumptions that are in place and that will be addressed by follow-up changes, but the goal is really to get the ball rolling behind a feature flag. I would encourage reviewers to not focus on edge cases but rather on the overall design of how this experimental feature is implemented.

When the experimental feature that this PR provides is enabled, it depends on changes in several other repositories like node-sdc-clients, a new sdc-volapi repository, sdcadm, mountain-gorilla, etc. but they should essentially have no impact when the feature is disabled, which is the default.

Tests that have been added are written to pass when the experimental feature is disabled. Thus, these changes could land independently of the rest of the changes needed to support NFS shared volumes.


docs/api/features/volumes.md
```diff
@@ -17,3 +17,36 @@ different from Docker Inc's docker:
...
+```
+
+This command will create a new core zone that runs the VOLAPI service, which
+implements the Volumes API. It will also enable the
+`experimental_nfs_shared_volumes` metadata property in SAPI.
+
+At this point, all `docker volume` commands are supported but only for the
+`'tritonnfs'` volume driver, which provies support for NFS shared volumes. Note
+that the `'tritonnfs'` volume driver needs to be specified in the `docker volume
+create` command for it to work.
```
joshwilsdon: How does the default volume driver work? Is it possible for us to eventually make this the default driver? Or will users always have to specify one?

lib/backends/sdc/containers.js
```diff
@@ -35,7 +35,7 @@ var images = require('./images');
 var Link = require('../../models/link');
 var utils = require('./utils');
 var validate = require('../../validate');
-
+var networks = require('../../networks');
```
joshwilsdon: NICE!

lib/backends/sdc/containers.js
```diff
@@ -1676,20 +1577,22 @@ function buildVmPayload(opts, container, callback) {
...
-        || container.HostConfig && container.HostConfig.Binds;
-    if (binds && binds.length > 0) {
-        log.error({host_volumes: binds},
-            'host volumes are not supported');
-        callback(new errors.DockerError(
-            'host volumes are not supported'));
-        return;
+    if (opts.config.experimental_nfs_shared_volumes !== true) {
+       /*
+        * "Host" volumes created using -v http[s]://example/file:/container/file
```
joshwilsdon: This comment looks leftover from the old implementation of hostvolumes. Probably just delete it and replace with something about how we don't support host volumes but we do experimentally support NFS volumes which use the host volume syntax?

misterdjules: This code runs when experimental NFS shared volumes is _not_ enabled, and the goal was to provide the exact same behavior than today in that case. Does that make sense?

docs/api/features/volumes.md
misterdjules: Good point, RFD states:

> The tritonnfs driver is the default driver on Triton

whereas my thought was that in order to avoid confusion between docker/dockers' default volume driver and a default driver with a different behavior, we'd require users to specify `tritonnfs`. If that doesn't make sense, I'll make `tritonnfs` be the default driver on Triton.

Thanks for bringing that up!

lib/backends/sdc/containers.js
joshwilsdon: Yeah, that part makes sense. I was just pointing out that this comment is already wrong in master. It's fine to just fix that separately too.

lib/backends/sdc/containers.js
```diff
@@ -1846,6 +1749,66 @@ function buildVmPayload(opts, container, callback) {
...
+                return;
+            }
+
+            log.debug('handling shared volumes');
+
+            binds = container.Binds
+                || container.HostConfig && container.HostConfig.Binds;
+
+            if (binds && binds.length > 0) {
+                var volumeComponents = binds[0].split(':');
```
joshwilsdon: A comment here explaining what this "binds" or "binds[0]" are actually expected to look like might be helpful. We're also ignoring options for now too, right? E.g. -v myvolume:/mymnt:ro?

lib/backends/sdc/containers.js
misterdjules: Ah OK, I understand. I might file another PR for that, thanks!

lib/backends/sdc/containers.js
misterdjules: > A comment here explaining what this "binds" or "binds[0]" are actually expected to look like might be helpful.

Absolutely, I will add a comment.

> We're also ignoring options for now too, right? E.g. -v myvolume:/mymnt:ro?

For now yes, this is part of the follow-up tickets/change that I'll file when/if that PR gets merged.

lib/backends/sdc/volumes.js
```diff
@@ -0,0 +1,156 @@
...
+    var volapiClient = options.app.volapi;
+
+    params.owner_uuid = options.account.uuid;
+
+    volapiClient.listVolumes(params, {headers: {'x-request-id': options.reqId}},
+        function onVolumesListed(err, volumes) {
+            log.debug({err: err, volumes: volumes},
+                'Response from volapi ListVolumes');
+            if (err) {
+                callback(errors.volapiErrorWrap(err, 'problem list volumes'));
```
joshwilsdon: nit: "problem list**ing** volumes"

lib/endpoints/volumes.js
```diff
@@ -0,0 +1,247 @@
...
+        app: req.app,
+        log: log,
+        reqId: req.getId()
+    };
+
+    req.backend.inspectVolume({
+        name: req.params.name
+    }, options, function onVolumeInspected(err, volume) {
+        if (err) {
+            log.error({err: err}, 'Error when deleting volume');
```
joshwilsdon: "Error when **inspecting** volume"?

package.json
```diff
@@ -20,7 +20,7 @@
     "openssl-wrapper": "0.2.1",
     "restify": "git://github.com/joyent/node-restify.git#fd5d5b5",
     "sdc-bunyan-serializers": "git+ssh://git@github.com:joyent/sdc-bunyan-serializers.git#5454ae4",
-    "sdc-clients": "~9.1",
+    "sdc-clients": "git+ssh://git@github.com:joyent/node-sdc-clients.git#f1633e0aa83b9dca1e844c62fa10f20476cca1af",
```
joshwilsdon: can we use https?

test/integration/cli-nfs-shared-volumes.test.js
```diff
@@ -0,0 +1,236 @@
...
+                checkVolumesSupportDisabled(t, err, stderr);
+            }
+
+            t.end();
+        });
+    });
+
+    // Skip next tests if NFS shared volumes are not supported, as they would
+    // fail since the volume that needs to be mounted would have not been
+    // created. Testing that mounting an inexisting module fails is done in a
```
joshwilsdon: s/inexisting/nonexistent/

test/integration/cli-nfs-shared-volumes.test.js
```diff
@@ -0,0 +1,236 @@
...
+    // separate test, and this allows us to make the rest of this test slighly
```
joshwilsdon: s/slighly/slightly/

joshwilsdon:
I read through this PR and I think overall it looks good. There were a few very minor issues I pointed out, but since I know you were more concerned with the overall structure / etc, I think this looks good.

package.json
misterdjules: Yes, will fix.

test/integration/cli-nfs-shared-volumes.test.js
misterdjules: Thanks, will fix.

test/integration/cli-nfs-shared-volumes.test.js
misterdjules: Thanks, will fix.

lib/endpoints/volumes.js
misterdjules: Will fix.

lib/backends/sdc/volumes.js
misterdjules: Will fix.

misterdjules:
@joshwilsdon Added 3 commits to this PR that I believe address all your comments.

joshwilsdon:
lgtm

WyseNynja:
This looks really handy. Is there a timeline for getting this merged in?

misterdjules:
It was decided that the whole feature needed to be more robust across all impacted APIs in Triton before merging anything into master branches of any Triton repository. As a result, development for the support of shared NFS volumes has continued on feature branches named `tritonnfs` for all relevant Triton repositories.

One way to test this feature is to [run the `setup.sh` script from the sdc-volapi repository](https://github.com/joyent/sdc-volapi#installation). This will upgrade all the relevant services and tools in a given Triton DC with the images available in the experimental channel that are built from the `tritonnfs` branches of all the relevant repositories.

Moreover, given that the contribution process has changed from using GitHub to using our Gerrit service, I think it's time to close this PR.

Feel free to post more comments in here if you have more questions about the state of shared volumes support in Triton though, I should be able to answer them quickly.

Thank you for your interest and sorry for the confusion!

PR #79, opened 2016-04-19
DOCKER-788: update image/container filtering for new API format

twhiteman:
With Docker 1.10 (API version 1.22) the filters changed from an array to an
object with the value true.

In 1.9.1: 
```json
    filters={"label":["com.joyent.package=sample-2G", "foo=bar"]}
```

In 1.10.0:
```json
    filters={"label":{"com.joyent.package=sample-2G":true, "foo=bar":true}}
```

### DIFF
--- a/lib/backends/sdc/containers.js
+++ b/lib/backends/sdc/containers.js
@@ -2321,12 +2321,12 @@ function getContainers(opts, callback) {
 
     if (opts.filters) {
         filters = JSON.parse(opts.filters);
-        filters = utils.getNormalizedFilters({ log: log }, filters);
+        filters = utils.getNormalizedFilters(filters);
         if (filters instanceof Error) {
             callback(new errors.DockerError('invalid filters: ' + filters));
             return;
         }
-        log.debug('filters: ', filters);
+        log.debug({filters: filters}, 'getContainers: filters');
     }
 
     vasync.pipeline({arg: {}, funcs: [
--- a/lib/backends/sdc/containers.js
+++ b/lib/backends/sdc/containers.js
@@ -2309,13 +2309,25 @@ function getContainers(opts, callback) {
     assert.object(opts.app.vmapi, 'opts.app.vmapi');
     assert.number(opts.clientApiVersion, 'opts.clientApiVersion');
     assert.arrayOfObject(opts.images, 'opts.images');
+    assert.optionalString(opts.filters, 'opts.filters');
     assert.optionalObject(opts.log, 'opts.log');
     assert.string(opts.req_id, 'opts.req_id');
 
+    var filters;
     var log = opts.log || this.log;
     var pkgmapNtoU = {};
     var pkgmapUtoN = {};
     var vmapi = opts.app.vmapi;
+
+    if (opts.filters) {
+        filters = JSON.parse(opts.filters);
+        filters = utils.getNormalizedFilters({ log: log }, filters);
+        if (filters instanceof Error) {
+            callback(new errors.DockerError('invalid filters: ' + filters));
+            return;
+        }
+        log.debug('filters: ', filters);
+    }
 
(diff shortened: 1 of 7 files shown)


lib/backends/sdc/containers.js
```diff
@@ -2309,14 +2309,26 @@ function getContainers(opts, callback) {
...
     var vmapi = opts.app.vmapi;
 
+    if (opts.filters) {
+        filters = JSON.parse(opts.filters);
+        filters = utils.getNormalizedFilters({ log: log }, filters);
+        if (filters instanceof Error) {
+            callback(new errors.DockerError('invalid filters: ' + filters));
+            return;
+        }
+        log.debug('filters: ', filters);
```
trentm: Still want this? If so, please this:

`log.debug({filters: filters}, 'getContainers: filters');`

lib/backends/sdc/utils.js
```diff
@@ -110,45 +110,46 @@ function getPublishingRules(opts, fwrules) {
...
+        } else if (typeof (val) === 'object') {
+            // Convert to an array of keys.
+            newFilters[name] = Object.keys(val);
+        } else {
+            log.warn({
+                filters: filters,
+                type: typeof (val),
+                val: JSON.stringify(val)
+            }, 'WARNING: filter %s is not an array or object', name);
+            return new Error('invalid filters');
```
trentm: I'd drop the `log.warn` (I'm assuming we are going to raise the error for the response and then the audit log will show it).

Please have the Error message include details:  `return new Error(format('invalid filters: filter is not an array or object: %j', val));`

lib/backends/sdc/utils.js
```diff
@@ -110,45 +110,46 @@ function getPublishingRules(opts, fwrules) {
...
  * In 1.10.0:
  *
- *     filters={"label":{"com.joyent.package=sample-2G":true}}
- *
- * This function takes the filters[label] value (as the filters parameter here)
- * and returns a normalized array of the packages we want to filter on.
+ *     filters={"label":{"com.joyent.package=sample-2G":true, "foo=bar":true}}
  *
+ * This function takes the filters object and returns a new normalized object
+ * with each object value being an array instance.
```
trentm: I.e. the normalized form is  what it looks like in 1.9.1, right?  Might be clearer to state that or just say "the former".

test/lib/common.js
```diff
@@ -251,6 +251,59 @@ function makeContainerName(prefix) {
 }
 
 
+/*
+ * Parse docker columnar output by using the width's in the first header row.
```
trentm: "widths"

trentm:
Were there tests already for filtering on *container* labels? I see one test for filtering on image labels.

trentm:
I assume this passes the suite with each of the CLI versions (including 1.11).

LGTM.

lib/backends/sdc/containers.js
twhiteman: Done

lib/backends/sdc/utils.js
twhiteman: Done

lib/backends/sdc/utils.js
twhiteman: Done

test/lib/common.js
twhiteman: Done

twhiteman:
There were no tests in sdc-docker covering container filtering - but there is good coverage in the docker/docker nightly tests.

Since the docker/docker integration tests are currently strapped to one version of docker (1.9.2 at time of writing), this is why we didn't notice this breakage.

I'll add a few cli container filter tests to sdc-docker, so that we can have *some* coverage of this across multiple docker versions.

twhiteman:
Added container filter tests in new 'cli-filters.test.js' file.

PR #80, opened 2016-04-19
DOCKER-630: Allow user to pass locality hint/restriction in docker run/create (swarm-iness)

trentm:


lib/backends/sdc/affinity.js
```diff
@@ -0,0 +1,722 @@
...
+var errors = require('../../errors');
+
+
+// ---- globals
+
+var FILTER_KEY_RE = /^[a-z_][a-z0-9\-_.]+$/i;
+// Note that this differs from the Swarm one in expr.go to fix some issues
+// and accomodate slight parsing differences.
+// JSSTYLED
+var FILTER_VALUE_RE = /^[-a-z0-9:_\s.*/()?+[\]\\^$|]+$/i;
```
joshwilsdon: Might be good to list which types of values this is supposed to allow. Regexes are not the most readable.

test/unit/affinity.test.js
```diff
@@ -0,0 +1,726 @@
...
+    var o = obj;
+    var parts = lookup.split('.');
+    var s = [];
+    for (var i = 0; i < parts.length; i++) {
+        var part = parts[i];
+        if (part === 'tag') {
+            part = 'tags';  // VMAPI "?tag.foo=bar" looks up "tags.foo".
+        }
+        s.push(part);
+        if (!hasOwnProperty.call(o, part)) {
```
joshwilsdon: To me:

```
if (!o.hasOwnProperty(part)) {
```

is much more readable.

tools/jsl.node.conf
```diff
@@ -122,6 +122,7 @@
 +define setInterval
 +define setTimeout
 +define setImmediate
++define hasOwnProperty
```
joshwilsdon: This is unneeded if you use o.hasOwnProperty() as suggested at L53 of test/unit/affinity.test.js

lib/backends/sdc/affinity.js
```diff
@@ -0,0 +1,736 @@
...
+            if (err && err.statusCode !== 404) {
+                cb(err);
+            } else if (!err && vm_ && vm_.docker
+                && ['destroyed', 'failed'].indexOf(vm_.state) === -1
+                && vm_.internal_metadata['docker:id'] === aff.value)
+            {
+                /*
+                 * If the 'docker:id' check fails, treat this like a 404.
+                 * This could happen if someone made up a docker_id that
+                 * started with the 32-chars for a real VM.
```
joshwilsdon: This block comment doesn't look like it's describing the code here. Does that belong in the else?

trentm:
squash merged this branch to master.

PR #82, opened 2016-04-21
DOCKER-814: placement/affinity docs

trentm:
@misterbisson @askfongjojo  Care to review this?

### DIFF
--- a/docs/api/commands/create.md
+++ b/docs/api/commands/create.md
@@ -161,6 +161,10 @@ Triton's secure, multi-tenant, container-native environment imposes some differe
 * `-v`, `--volume` and `--volumes-from` behave slightly differently in a Triton's container-native environment. See [volumes](../features/volumes.md).
 * `--volume-driver` and other `docker volume` commands are unimplemented at this time.
 
+Docker Swarm's affinity filters for controlling on which server a container
+is provisioned are supported. See the [placement feature
+documentation](../features/placement.md) for details.
+
 Please contact Joyent support or file a ticket if you discover any additional divergence.
 
 ## Related
--- a/docs/api/features/placement.md
+++ b/docs/api/features/placement.md
@@ -50,7 +50,7 @@ is exposed via the `compute_node` field using the [Triton CLI](XXX):
 to talk to a pool of Docker hosts (a.k.a. nodes) as a single virtual Docker
 host. `docker run` against a Swarm master will choose a node on which to run
 your container. Sometimes it matters which node is selected -- my
-webhead should run near the redis is uses for caching, this database instance
+webhead should run near the redis it uses for caching, this database instance
 should *not* be on the same node as other database instances in the HA cluster.
 Swarm defines "filters" for node selection, and in particular ["affinity
 filters"](https://docs.docker.com/swarm/scheduler/filter/#use-an-affinity-filter)
--- a/docs/api/features/placement.md
+++ b/docs/api/features/placement.md
@@ -1,3 +1,137 @@
 # Placement
 
-What are the default rules in SDC, do they differ in hosted Triton?
+When using Docker with a Triton data center (a.k.a. SDC) as your Docker Host,
+your Docker containers can be placed on any physical server in the data center.
+This is one of the basic benefits of Triton's Docker solution: with just the
+vanilla `docker` client you can get the benefit of spreading your containers
+across multiple servers, setting the groundwork for high availability (HA)
+services and avoiding single points of failure (SPOF).
+
+This document explains how Triton places new containers on servers and what
+facilities are exposed for controlling placement.
+
+
+## Default placement
+
+By default, Triton attempts to spread all containers (and non-Docker containers
+and VMs) owned by a single account across separate physical servers.
+
+Within a Docker container the physical server on which a container is running
+is exposed via the [`sdc:server_uuid` metadata
+key](http://eng.joyent.com/mdata/datadict.html):
+
+    $ docker run -ti --rm alpine /bin/sh
+    / # hostname
+    cc72c36cb591
+    / # /native/usr/sbin/mdata-get sdc:server_uuid
+    44454c4c-5400-1034-8052-b5c04f383432
+
+By running another container we can see this default spread behavior (the
+`server_uuid` differs):
+
+    $ docker run -ti --rm alpine /bin/sh
+    / # hostname
+    98d5a22a0c5e
+    / # /native/usr/sbin/mdata-get sdc:server_uuid
+    44454c4c-4400-1054-8052-b5c04f383432
+
+Outside the containers, the physical server on which a container is running
+is exposed via the `compute_node` field using the [Triton CLI](XXX):
+
+    $ triton insts -o shortid,name,age,compute_node
+    SHORTID   NAME             AGE  COMPUTE_NODE
+    cc72c36c  serene_bose      6m   44454c4c-5400-1034-8052-b5c04f383432
+    98d5a22a  goofy_engelbart  2m   44454c4c-4400-1054-8052-b5c04f383432
+
+
+## Swarm affinity
+
+[Docker Swarm](https://docs.docker.com/swarm/overview/) is a system to allow you
+to talk to a pool of Docker hosts (a.k.a. nodes) as a single virtual Docker
+host. `docker run` against a Swarm master will choose a node on which to run
+your container. Sometimes it matters which node is selected -- my
+webhead should run near the redis is uses for caching, this database instance
+should *not* be on the same node as other database instances in the HA cluster.
+Swarm defines "filters" for node selection, and in particular ["affinity
+filters"](https://docs.docker.com/swarm/scheduler/filter/#use-an-affinity-filter)
+where node selection is described in terms of existing containers. For example:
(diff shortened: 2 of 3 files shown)


docs/api/features/placement.md
```diff
@@ -1,3 +1,137 @@
...
+    98d5a22a  goofy_engelbart  2m   44454c4c-4400-1054-8052-b5c04f383432
+
+
+## Swarm affinity
+
+[Docker Swarm](https://docs.docker.com/swarm/overview/) is a system to allow you
+to talk to a pool of Docker hosts (a.k.a. nodes) as a single virtual Docker
+host. `docker run` against a Swarm master will choose a node on which to run
+your container. Sometimes it matters which node is selected -- my
+webhead should run near the redis is uses for caching, this database instance
```
misterbisson: > near the redis is uses for caching

?

misterbisson:
One part that's unclear, otherwise lgtm

docs/api/features/placement.md
trentm: @misterbisson Typo. "near the redis it uses for caching" ... is that clear then?

misterbisson: works for me

docs/api/features/placement.md
```diff
@@ -1,3 +1,137 @@
...
+services and avoiding single points of failure (SPOF).
+
+This document explains how Triton places new containers on servers and what
+facilities are exposed for controlling placement.
+
+
+## Default placement
+
+By default, Triton attempts to spread all containers (and non-Docker containers
+and VMs) owned by a single account across separate physical servers.
```
askfongjojo: Hrm only east3b uses random spread strategy. The rest of JPC is still on min memory.

trentm: actually I think the locality spread might "win" here.

trentm: Yup, locality spread for an account wins. E.g. in east1:

```
$ for i in {0..9}; do docker run --name spread$i -d alpine sleep 3600 & done

# Then later:
$ triton insts -o shortid,name,state,compute_node,flags,age
SHORTID   NAME     STATE    COMPUTE_NODE                          FLAGS  AGE
9676a655  spread4  running  44454c4c-4400-1059-804e-b5c04f383432  DF     1m
4a55227a  spread8  running  44454c4c-4400-1052-8053-b5c04f383432  DF     1m
c0660b29  spread9  running  44454c4c-4400-1052-8050-b5c04f383432  DF     1m
b9627903  spread2  running  44454c4c-4400-1032-8051-b5c04f383432  DF     1m
37d9f2ff  spread3  running  44454c4c-4400-1057-804d-b5c04f383432  DF     1m
99886144  spread6  running  44454c4c-4400-1034-8051-b5c04f383432  DF     59s
4454641c  spread1  running  44454c4c-4400-1051-8052-b5c04f383432  DF     55s
bd96884f  spread0  running  44454c4c-4400-1056-8053-b5c04f383432  DF     48s
9a2e9d59  spread5  running  44454c4c-4400-1059-804d-b5c04f383432  DF     45s
2356b445  spread7  running  44454c4c-5400-1044-804d-b5c04f383432  DF     39s

$ triton insts -Ho compute_node | sort | uniq -c
   1 44454c4c-4400-1032-8051-b5c04f383432
   1 44454c4c-4400-1034-8051-b5c04f383432
   1 44454c4c-4400-1051-8052-b5c04f383432
   1 44454c4c-4400-1052-8050-b5c04f383432
   1 44454c4c-4400-1052-8053-b5c04f383432
   1 44454c4c-4400-1056-8053-b5c04f383432
   1 44454c4c-4400-1057-804d-b5c04f383432
   1 44454c4c-4400-1059-804d-b5c04f383432
   1 44454c4c-4400-1059-804e-b5c04f383432
   1 44454c4c-5400-1044-804d-b5c04f383432
```

askfongjojo: Yes that's as expected, because dogpile happens after the one round of spread is there. With east-1, there are a larger pool of CNs but if there is a smaller DC, and the no. of containers the account has > the number of available CNs, it won't be spreading when every CN has one instance for that account. The subsequent ones will be bin-packed.

docs/api/features/placement.md
```diff
@@ -1,3 +1,137 @@
...
+`<op>` is one of:
+
+- `==`: The new container must be on the same node as the container(s)
+  identified by `<value>`.
+- `!=`: The new container must be on a different node as the container(s)
+  identified by `<value>`.
+- `==~`: The new container should be on the same node as the container(s)
+  identified by `<value>`. I.e. this is a best effort or "soft" rule.
+- `!=~`: The new container should be on a different node as the container(s)
+  identified by `<value>`. I.e. this is a best effort or "soft" rule.
```
askfongjojo: There is a divergence - currently hard and soft requirements cannot be mixed in the same provisioning request. This is mentioned in DOCKER-630 and what I observed as well. Triton only derives the soft vs hard flag based on the first condition. E.g. when there is container!=foo and container =~bar in the request, the requirements are both evaluated as "hard".

docs/api/features/placement.md
```diff
@@ -1,3 +1,137 @@
...
+    docker run -e 'affinity:container!=foo*' ...
+
+    # Same, using a regular expression:
+    docker run -e 'affinity:container!=/^foo/' ...
+
+Note: At the time of writing the label syntax is in Docker Swarm
+[code](https://github.com/docker/swarm/blob/d9beef7/cluster/config.go#L83-L86),
+but not [documented](https://docs.docker.com/swarm/scheduler/filter/).
+
+
```
askfongjojo: Maybe we want to mention what to expect when a request cannot be fulfilled based on the affinity requirements (though the error is not the best at the moment). Something like this?

If no suitable server can be located based on the affinity requirements, the provisioning request will fail with the following error message:
    (DockerNoComputeResourcesError) No compute resources available.

docs/api/commands/create.md
```diff
@@ -161,6 +161,10 @@ Triton's secure, multi-tenant, container-native environment imposes some differe
 * `-v`, `--volume` and `--volumes-from` behave slightly differently in a Triton's container-native environment. See [volumes](../features/volumes.md).
 * `--volume-driver` and other `docker volume` commands are unimplemented at this time.
 
+Docker Swarm's affinity filters for controlling on which server a container
```
tgross: Looks like we're manually breaking lines here rather than letting the markdown rendering wrap them.

trentm: @tgross I consistently wrap at <80 columns. ;)  That, IMO, helps with developers reading/editing the raw Markdown docs; and with diffs.  I find the crazy-long lines frustrating to edit.

It *does* require the author to be using an editor that can do paragraph reflow.

docs/api/features/placement.md
trentm: @askfongjojo From my recent updates to the DAPI algorithm here (https://github.com/joyent/sdc-designation/blob/master/lib/algorithms/soft-filter-locality-hints.js#L63):

```
 * - default: If no "locality" is given, this algoritm does a non-strict
 *   attempt to filter out servers with VMs owned by this customer. In other
 *   words, by default Triton attempts to spread out a customer's VMs.
 *   Customer-facing documentation should promise no more than Triton making
 *   a reasonable attempt to spread out customer VMs by default -- i.e.
 *   exactly how should be left under-specified to allow for future changes.
 *   (Dev Note: Future changes might be: better balancing, balancing by
 *   RBACv2 project, balancing by image, limiting the number of CNs a
 *   customer can land on. This balancing by default *overrides* subsequent
 *   sorting/scoring attempts in the sdc-designation pipeline.)
```

This is why I used "attempts" in the language above. But perhaps that still gives the wrong impression to a user that may have lots of containers. I'll try to give a more accurate expectation there.

I also wouldn't mind making a quick attempt at an improvement here: to have the default account-spread in the locality algorithm not totally give up if the customer has one VM on every server. A decent attempt at that might be best to wait for the "score"ing stuff that Marsell is working on in DAPI-298/DAPI-301. E.g. we'd bias slightly in favour of CNs without a lot of the account's VMs. Come RBACv2 we'll probably want to do it by project anyway.

docs/api/features/placement.md
trentm: Thanks, yes I'll add a note about this.

> E.g. when there is "container!=foo" and "container==~bar" in the request, the requirements are both evaluated as "hard".

Actually given both hard and soft affinities *the soft ones are ignored*: https://github.com/joyent/sdc-docker/blob/f1684e3f32335055986f1e47381d3494d070c506/lib/backends/sdc/affinity.js#L672-L677

docs/api/features/placement.md
trentm: Yes, good point. I'll add that.

trentm:
@askfongjojo Added commit  	22eb7d1 for your feedback.

docs/api/features/placement.md
askfongjojo: Ah ok. I didn't try all the permutations. Hard one wins - that makes sense.

trentm:
merged in commit d32f66162e99a6788f80bdf822dfae7646226470

PR #87, opened 2016-07-12
WIP: DOCKER-585

notmatt:


lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+ *     - a name is preferred to a partial id
+ *     - a full id is preferred to a name
  */
 function addNetworksToPayload(opts, container, payload, callback) {
     assert.object(opts, 'opts');
     assert.object(opts.config, 'opts.config');
     assert.object(opts.config.napi, 'opts.config.napi');
     assert.object(opts.log, 'opts.log');
+
+    assert.object(opts.config.overlay);
```
misterdjules: Does `assert-plus` need a second string parameter to produce a meaningful assertion message, like with the assert statements above?

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
-        opts.log.debug('Fabrics configured: using for networking');
-        return addFabricNetworksToPayload(opts, container, payload, callback);
+    /*
+     * currently there are four network possibilities:
+     *  - no fabrics configured -> external network.
+     *  - fabrics configured and...
+     *    - 'bridge' or no network specified: default fabric and...
+     *      - if publishing ports or building: also external network
+     *    - fabric name specified: named fabric only
+     *    unsupported:
```
misterdjules: It's not clear to me what's "unsupported".

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+     *
+     * We need to determine our network 'mode'. Docker defines the following
+     * networking modes:
+     *
+     * 1. 'legacy bridge' - the default currently supported.
+     * 2. 'host' - not meaningful for us
+     * 2. 'container' - like 'host', not meaningful
+     * 3. 'none' - eventually supportable
+     * 4. 'user-defined bridge' - probably supportable
+     * 5. 'user-defined overlay' - definitely supportable
```
misterdjules: Is `'user-defined overlay'` supportable or supported? From my understanding of the comments above, I would think that it's supported, not only supportable.

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+/*
+ * When fabrics are not configured, we will use the named
+ * network, defaulting to 'external'.
+ */
+function externalNetworkByName(opts, container, payload, callback) {
+    assert.object(opts.config, 'opts.config');
+    assert.optionalString(opts.config.externalNetwork,
+        'opts.config.externalNetwork');
+    assert.object(opts.account, 'opts.account');
+    assert.string(opts.account.uuid, 'opts.account.uuid');
```
misterdjules: Do we also want to `assert` on the other parameters?

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
-            callback(errors.napiErrorWrap(err, 'problem listing networks'));
+            callback(errors.napiErrorWrap(err,
+                format('Networks: problem listing network %s',
+                    listParams.name)));
+            return;
+        }
+
+        if (networks.length < 1) {
+            callback(new errors.DockerError(
+                format('unable to find %s network uuid', listParams.name)));
```
misterdjules: nit: is `listParams.name` a network _uuid_, or can it be either a name or a uuid? In that case do we want to not specify "uuid" in the error message?

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+ * or if 'bridge' is supplied, we will use the user's default fabric
+ * network, stored in UFDS.
+ *
+ * Additionally, if the user is publishing ports or we have the
+ * `fabricRequireExternal` option (currently used only by docker build,
+ * see DOCKER-705), we will also attach to configured network pool
+ * (typically the external/public pool).
+ */
+function defaultFabricNetwork(opts, container, payload, callback) {
+    var dc = opts.app.config.datacenterName;
```
misterdjules: I would suggest `assert`ing on all parameters.

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
-        networks.forEach(function (n) {
-            if (!external_net
-                && n.name === netName) {
+        payload.networks = [ {uuid: conf.defaultnetwork} ];
+
+        if (requireExternal) {
+            payload.networks.push({uuid: opts.config.overlay.externalPool});
+        }
+
+        payload.networks[payload.networks.length - 1].primary = true;
```
misterdjules: Instead of making the last network to be added to the networks as primary, would it better to set the primary attribute specifically on either the external network? Something like:

```
if (requireExternal) {
    externalNetwork = {
      uuid: opts.config.overlay.externalPool,
      primary: true
    };
   
    payload.networks.push(externalNetwork);
}
```

It seems it would be more robust if the order of creation of network objects changes, or if a new network object gets added to the list of network as part of that function's implementation.

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
 
-                external_net = n.uuid;
+    vasync.parallel({
+        funcs: [
+            function byName(cb) {
+                var listParams = {
+                    name: query,
+                    fabric: true,
+                    owner_uuid: opts.account.uuid,
+                    limit: 2
```
misterdjules: Why is limit set to `2`?

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+            function byDockerId(cb) {
+                // to search for docker ids, when NAPI only retains UUIDs, we
+                // rely on the 'double uuid' convention (see
+                // utils.networkUuidToDockerId()). To perform the search,
+                // we truncate the provided string to 8 chars, perform a
+                // wildcard search on networks, and post-filter the results to
+                // find more-complete matches than the at-most-8-char prefix
+                // might find.
+
+                var queryString = utils.ldapEscape(query.substr(0, 8)) + '*';
```
misterdjules: Would `uuidSearchPattern` or something along those lines be a better name instead of `queryString`? It seems this variable does not represent the whole querystring of the following `listNetworks` call/request.

misterdjules: Why not searching using the full UUID (potentially reformatted if need be to not contain `-` characters), or at least a UUID as long as a NAPI network UUID?

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+        log.debug({ networks: networks },
+            'Networks: found candidate networks, sorting');
+
+        // pick the 'best' match, following Docker's heuristic that:
+        // 1. exact id is always best
+        // 2. exact name match
+        // 3. partial id match
+        // >1 partial id is an ambiguous result, but we do not know
+        // the order in which we're going to evaluate networks.
+        var network = networks.reduce(function (acc, nw, i, c) {
```
misterdjules: I find the short names for parameters `nw`, `i` and `c` unclear, is it possible to use longer/clearer names?

misterdjules: In the same idea, maybe `acc` should be renamed `currentBestMatch` to better convey how it's used?

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,302 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+            var exactId = utils.networkUuidToDockerId(nw.uuid) === query;
+            var exactName = nw.name === query;
+
+            if (exactId) {
+                // always wins.
+                acc = nw;
+            } else if (exactName && !idMatch) {
+                // wins if we don't have an id match.
+                acc = nw;
+            } else if (acc && !idMatch && !nameMatch) {
```
misterdjules: Isn't `acc` always guaranteed to evaluate to a truthy value?

misterdjules: My apologies, `acc` is undefined unless a second parameter is passed to `reduce`.

lib/backends/sdc/utils.js
```diff
@@ -942,16 +946,26 @@ function compressPorts(ports) {
     return ranges;
 }
 
+/*
+ * Escapes strings for ldap filters.
+ * See: pfmooney/node-ldap-filter/lib/substr_filter.js#L12-L15
+ */
+function ldapEscape(str) {
+    /* JSSTYLED */
+    return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
```
misterdjules: Is there a way to get that from the `ldap-filter` module and not duplicate it here?

lib/validate.js
```diff
@@ -226,6 +226,53 @@ function assertGelfAddress(arg, name) {
...
+ *
+ * We support: bridge, other.
+ */
+function assertNetworkMode(networkmode, name) {
+    if (networkmode === 'bridge') {
+        return;
+    }
+
+    if (networkmode === 'host' || networkmode === 'none') {
+        throw new errors.ValidationError(fmt('NetworkMode %s is not supported',
```
misterdjules: Would [NotImplementedError](https://github.com/joyent/sdc-docker/blob/master/lib/errors.js#L256-L268) be better suited for such errors?

test/integration/api-create.test.js
```diff
@@ -90,250 +93,481 @@ test('setup', function (tt) {
...
+
+test('ensure fabrics enabled', function (tt) {
+    tt.test('fabric configuration', function (t) {
+        var listOpts = {};
+        var listParams = {};
+        NAPI.listFabricVLANs(ALICE.account.uuid, listOpts, listParams,
+            function (err, vlans) {
+                if (err) {
+                    FABRICS = false;
+                    if (err.code !== 'PreconditionRequiredError') {
```
misterdjules: What does `'PreconditionRequiredError'` represent? Are fabrics enabled in nightly or other DCs where tests are run automatically?

test/integration/api-create.test.js
```diff
@@ -90,250 +93,481 @@ test('setup', function (tt) {
...
     });
 
-    tt.test('docker rm', function (t) {
-        DOCKER_ALICE.del('/containers/' + created.id, ondel);
-
-        function ondel(err, res, req, body) {
-            t.ifErr(err, 'rm container');
+    tt.test('fabric network setup', function (t) {
+        var nw1uuid = libuuid.create();
+        var nw2uuid = nw1uuid.replace(/-/g, '').substr(0, 12);
```
misterdjules: Would it be worth it to add a comment here that the UUID of the second network is meant to be created in a way that exercises the code that disambiguate network IDs, short IDs and names?

test/integration/api-create.test.js
```diff
@@ -90,250 +93,481 @@ test('setup', function (tt) {
...
+            name: 'alicetest1',
+            subnet: '10.0.8.0/24',
+            provision_start_ip: '10.0.8.2',
+            provision_end_ip: '10.0.8.254',
+            uuid: nw1uuid,
+            gateway: '10.0.8.1',
+            resolvers: ['8.8.8.8', '8.8.4.4']
+        };
+
+        // Must ignore the common naming convenetion for the sake of testing.
```
misterdjules: Nit: typo in `convenetion`.

test/integration/helpers.js
```diff
@@ -1637,6 +1652,43 @@ function didRestifyHandlerRun(reqId, handlerName, callback) {
     });
 }
 
+/*
+ * Gets or creates a fabric VLAN for use in testing.
+ */
+function getOrCreateFabricVLAN(client, userUuid, fabricParams, callback) {
+    client.getFabricVLAN(userUuid, fabricParams.vlan_id, {},
```
misterdjules: I think `assert`ing on all parameters would be useful.

test/integration/helpers.js
```diff
@@ -1637,6 +1652,43 @@ function didRestifyHandlerRun(reqId, handlerName, callback) {
...
+        }
+    );
+}
+
+/*
+ * Gets or creates a fabric network for use in testing; based on the
+ * network *name*.
+ */
+function getOrCreateFabricNetwork(client, userUuid, vlan_id, params, callback) {
+    var listParams = {
```
misterdjules: I think `assert`ing on all parameters would be useful.

lib/backends/sdc/containers.js
notmatt: I added an assert on `payload`, since that's the only one we write to; `container` is unused, and so I removed it from the parameters instead.

lib/backends/sdc/containers.js
notmatt: It's intended to be a name, and I meant to indicate that we couldn't determine the corresponding uuid - cleared up & switched to a NetworkNotFoundError.

lib/backends/sdc/containers.js
notmatt: Done as part of changes to support `-P` and `-p`

lib/backends/sdc/containers.js
notmatt: Removed; was left over from a previous version searching on a partial uuid (where >1 result would have been the user underspecifying). Anyway, removed here, not appropriate.

lib/backends/sdc/containers.js
notmatt: We aren't expecting a full UUID here, but a (potentially short/partial) docker id.

The second suggestion is a good one regardless; if we reformat the supplied id into a (potentially partial) uuid, we can remove the post-filtering step entirely.

lib/backends/sdc/utils.js
notmatt: Not without a new ldap-filter release, and maybe breaking its intended usage a bit. @pfmooney and I thought this was the best compromise.

There is a longer-term plan here, see [NAPI-367](https://smartos.org/bugview/NAPI-367), that should remove the requirement for any ldap filters on input, but right now this is required as a workaround (I'll note that too in next rev).

test/integration/api-create.test.js
notmatt: That's the error returned by NAPI when a fabric query is made, but fabrics aren't enabled.

Nightly does *not* have fabrics enabled at the moment.

lib/backends/sdc/containers.js
```diff
@@ -546,8 +546,7 @@ function addNetworksToPayload(opts, container, payload, callback) {
     assert.object(opts.config, 'opts.config');
     assert.object(opts.config.napi, 'opts.config.napi');
     assert.object(opts.log, 'opts.log');
-
-    assert.object(opts.config.overlay);
+    assert.object(opts.config.overlay, 'opts.config.overlay');
```
misterdjules: I usually assert on callback parameters too with `assert.func(callback, 'callback')`. That sometimes prevents me from spending too much time wondering why a sequence of async operations doesn't complete.

lib/backends/sdc/containers.js
```diff
@@ -593,12 +579,13 @@ function addNetworksToPayload(opts, container, payload, callback) {
...
  * network, defaulting to 'external'.
  */
-function externalNetworkByName(opts, container, payload, callback) {
+function externalNetworkByName(opts, payload, callback) {
     assert.object(opts.config, 'opts.config');
     assert.optionalString(opts.config.externalNetwork,
         'opts.config.externalNetwork');
     assert.object(opts.account, 'opts.account');
     assert.string(opts.account.uuid, 'opts.account.uuid');
+    assert.object(payload, 'payload');
```
misterdjules: Same as above regarding asserting on the `callback` parameter.

lib/backends/sdc/containers.js
```diff
@@ -665,13 +656,8 @@ function defaultFabricNetwork(opts, container, payload, callback) {
             return;
         }
 
-        payload.networks = [ {uuid: conf.defaultnetwork} ];
-
-        if (requireExternal) {
-            payload.networks.push({uuid: opts.config.overlay.externalPool});
-        }
+        payload.networks = [ { uuid: conf.defaultnetwork, primary: true } ];
```
misterdjules: Is it possible to end up with more than one primary network given that [the `primary` attribute can be set on the external network too](https://github.com/joyent/sdc-docker/pull/87/commits/bf5c44f5fbcac91f6cb590f8aa1d3d19329fbdbb#diff-d36e9a28d5d9b3702ee9fa48d5326b0dR1964)?

If so, is that a problem?

lib/backends/sdc/containers.js
```diff
@@ -681,43 +667,96 @@ function defaultFabricNetwork(opts, container, payload, callback) {
...
+    // 3. partial id match
     vasync.parallel({
         funcs: [
+            function byExactId(cb) {
+                // length === 64, and 0..31 === 32..63, or it's an 'impossible'
+                // id under our double-uuid convention, and we can skip it.
+                if (query.substr(0, 32) !== query.substr(32)) {
+                    log.debug({ query: query },
+                        'Networks: impossible exactId: %s, skipping', query);
+                    setImmediate(cb, null, []);
```
misterdjules: Should an error that means that the ID is invalid be passed to `cb`?

lib/backends/sdc/containers.js
```diff
@@ -681,43 +667,96 @@ function defaultFabricNetwork(opts, container, payload, callback) {
...
-                // we truncate the provided string to 8 chars, perform a
-                // wildcard search on networks, and post-filter the results to
-                // find more-complete matches than the at-most-8-char prefix
-                // might find.
+                // we assume the 'double uuid' convention for networks here,
+                // that is, dockerId = (uuid + uuid).replace(/-/g, '').
+                // So if we have a query.length > 31, 32... must be a prefix of
+                // 0..31. If not, the id supplied is impossible in our system.
+                if (query.length > 31
+                    && query.substr(0, query.substr(32).length)
```
misterdjules: if

```
query.substr(0, query.substr(32).length)
```

is used to get the first 32 characters of the string `query`, would

```
query.substr(0, 32)
```

be an equivalent but simpler way to achieve that?

misterdjules: Ah ok, I got it:

```
query.substr(0, query.substr(32).length)
```

is meant to get a string of same length as the second part of the Docker ID to test that that second part is a prefix of the first. I hadn't read the comment before reading the code.

lib/backends/sdc/containers.js
```diff
@@ -681,43 +667,96 @@ function defaultFabricNetwork(opts, container, payload, callback) {
...
+                    === query.substr(32)) {
+
+                    log.info({ query: query },
+                        'Networks: impossible network id query %s, skipping');
+                    setImmediate(cb, null, []);
```
misterdjules: Should an error be passed to `cb` in this case?

lib/backends/sdc/containers.js
```diff
@@ -681,43 +667,96 @@ function defaultFabricNetwork(opts, container, payload, callback) {
...
+                    return;
+                }
 
-                var queryString = utils.ldapEscape(query.substr(0, 8)) + '*';
+                // To perform the search, we transform the provided query to
```
misterdjules: Would it help readers to mention that the query itself can be a partial Docker ID, not only the result of the transformation from a Docker ID to a UUID?

lib/backends/sdc/containers.js
notmatt: I don't think so - the user input is ambiguously a name or a network ID, so in the general case we need to perform both searches. In the specific case where we can prove that the input can't possibly be a valid ID, we can skip the search. This doesn't mean it's an error, as we may still find a network name for the input.

lib/backends/sdc/containers.js
```diff
@@ -726,97 +765,68 @@ function namedNetwork(opts, container, payload, callback) {
...
+            // found result before an error, but did have errs.
+            log.warn({ err: err },
+                'Networks: non-critical error searching NAPI');
+        }
 
+        log.debug({ network: bestMatch.result }, 'Networks: chose %s/%s',
+            bestMatch.result.name, bestMatch.result.uuid);
 
-        payload.networks = [ { uuid: network.uuid, primary: true } ];
+        payload.networks = [ { uuid: bestMatch.result.uuid, primary: true } ];
```
misterdjules: Can the `primary` attribute be set on both this network _and_ on [the external network](https://github.com/joyent/sdc-docker/pull/87/commits/bf5c44f5fbcac91f6cb590f8aa1d3d19329fbdbb#diff-d36e9a28d5d9b3702ee9fa48d5326b0dR1964)? If so, is that a problem?

misterdjules: Ah, got it: not it's not possible because of https://github.com/joyent/sdc-docker/pull/87/commits/bf5c44f5fbcac91f6cb590f8aa1d3d19329fbdbb#diff-d36e9a28d5d9b3702ee9fa48d5326b0dR1959 :)

lib/backends/sdc/containers.js
misterdjules: Ah, got it: not it's not possible because of https://github.com/joyent/sdc-docker/pull/87/commits/bf5c44f5fbcac91f6cb590f8aa1d3d19329fbdbb#diff-d36e9a28d5d9b3702ee9fa48d5326b0dR1959 :)

lib/backends/sdc/containers.js
notmatt: See [above](https://github.com/joyent/sdc-docker/pull/87/files#r70852689)

lib/backends/sdc/utils.js
```diff
@@ -29,6 +29,34 @@ function networkUuidToDockerId(uuid) {
     return (uuid + uuid).replace(/-/g, '');
 }
 
+function shortNetworkIdToUuid(id) {
```
misterdjules: I think it would be helpful to add comments about what the valid inputs for this function are. Adding assertions could help to clarify/check which inputs are valid too.

For instance reading this code it wasn't clear to me whether `id` could be of any length, or was expected to have a length whose value was one of a set of values.

lib/backends/sdc/utils.js
```diff
@@ -29,6 +29,34 @@ function networkUuidToDockerId(uuid) {
...
+    var partial = '';
+
+    // consider a uuid as eight groups of four chars, plus hyphens.
+    // concatenate those groups in reverse order, jumping to the
+    // appropriate starting group.
+
+    switch (Math.ceil(id.substr(0, 32).length / 4)) {
```
misterdjules: Could this be achieved by using a regular expression similar to:

````
var matches = id.match(/([a-z0-9]{0,8})?([a-z0-9]{0,4})?.../);
````

and then iterate through from `matches[1]` through `matches.length` to reconstruct a UUID?

I feel like it could be easier to understand, but I realize this is subjective.

File: test/integration/api-create.test.js [hunk 29]
```diff
@@ -549,13 +588,38 @@ test('fabric setup', function (tt) {
...
+            dockerClient: DOCKER_ALICE,
+            test: t,
+            extra: { 'HostConfig.NetworkMode': fNetwork3.name },
+            start: true
+        }, oncreate);
+
+        function oncreate(err, result) {
+            var nics = result.vm.nics;
+            t.equal(nics.length, 1, 'only one nic');
+            t.equal(nics[0].network_uuid, fNetwork1.uuid, 'correct network');
```
misterdjules: If `fNetwork3` is named with the full dockerId of fNetwork1, and we prefer full network IDs, why isn't the container on the network `fNetwork3`? I feel that I'm missing something :)

notmatt: This is the pathological case where a user has taken the `id` of `fNetwork1` and used it as the `name` of `fNetwork3`. So if the user specifies that same string, docker (and hence us) should pick the network whose `id` matches, not the name.

lib/backends/sdc/containers.js
notmatt: Removed this comment in later revision; it's more concerned with planning for future support of docker features than with this code specifically.

lib/backends/sdc/containers.js
notmatt: Removed this comment in later revision; it's more concerned with planning for future support of docker features than with this code specifically.

lib/backends/sdc/containers.js
notmatt: Removed in latest changes!

lib/backends/sdc/containers.js
notmatt: You were correct the first time - `acc` should initialize to the first element of the array if it's not specified as the second param, and so was a bit redundant. But this whole filtering step has been replaced. :)

lib/backends/sdc/containers.js
notmatt: Done wherever I found them!

lib/backends/sdc/utils.js
```diff
@@ -25,6 +25,38 @@ function vmUuidToShortDockerId(uuid) {
     return uuid.replace(/-/g, '').slice(0, 12);
 }
 
+function networkUuidToDockerId(uuid) {
+    return (uuid + uuid).replace(/-/g, '');
+}
+
+function shortNetworkIdToUuid(id) {
```
melloc: I think that `shortNetworkIdToUuidPrefix` would be better, to make it clearer that it isn't converting to a full UUID.

lib/backends/sdc/utils.js
melloc: I personally find the switch easier to read than the full regex, but I think that both would make sense to the reader.

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,312 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+                    listParams.name, listParams.provisionable_by));
+            callback(new errors.NetworkNotFoundError(listParams.name));
+            return;
+        }
 
-                external_net = n.uuid;
-            }
-        });
+        payload.networks = [ { uuid: networks[0], primary: true} ];
+        return callback();
```
melloc: This can just be `callback()`.

lib/backends/sdc/containers.js
```diff
@@ -1735,6 +1953,21 @@ function buildVmPayload(opts, container, callback) {
...
 
+        function addExternalNet(_, cb) {
+            if (publishingPorts(container) || opts.fabricRequireExternal) {
+                // external must be primary.
+                payload.networks.forEach(function (nw) {
+                    delete nw.primary;
+                });
+
+                payload.networks.push(
+                    { uuid: opts.config.overlay.externalPool, primary: true });
```
melloc: `opts.config.overlay.externalPool` will always be defined if fabrics are enabled, right?

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,312 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
-            callback(new errors.DockerError(
-                'unable to find "'+netName+'" network uuid'));
+/*
+ * When fabrics are configured, but no specific network is supplied,
+ * or if 'bridge' is supplied, we will use the user's default fabric
+ * network, stored in UFDS.
+ *
+ * Additionally, if the user is publishing ports or we have the
+ * `fabricRequireExternal` option (currently used only by docker build,
+ * see DOCKER-705), we will also attach to configured network pool
```
melloc: I think that `(used by docker build, for example)` would be better, so that if others use it in the future, the parenthetical won't become incorrect.

melloc:
@notmatt I left some comments on things I noticed when I looked over this yesterday. Everything looks good to me!

lib/backends/sdc/containers.js
misterdjules: Ah OK, I thought this condition being true when that query is an invalid input, as in it's an error and we don't know what to do with it.

So is the condition:

```
(query.substr(0, 32) !== query.substr(32))
```

meant to filter out anything that is not a 64 characters Docker ID? If so, it seems it would have a few false negative, like if `query` was the empty string.

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,316 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+ * Additionally, if the user is publishing ports or we have the
+ * `fabricRequireExternal` option (currently used by docker build,
+ * see DOCKER-705), we will also attach to configured network pool
+ * (typically the external/public pool).
+ */
+function defaultFabricNetwork(opts, payload, callback) {
+    assert.object(opts, 'opts');
+    assert.object(opts.app, 'opts.app');
+    assert.object(opts.app.config, 'opts.app.config');
+    assert.object(opts.app.config.datacenterName,
```
melloc: This should actually be `assert.string`, otherwise this assertion will fail.

test/integration/helpers.js
```diff
@@ -1637,6 +1652,52 @@ function didRestifyHandlerRun(reqId, handlerName, callback) {
     });
 }
 
+/*
+ * Gets or creates a fabric VLAN for use in testing.
+ */
+function getOrCreateFabricVLAN(client, userUuid, fabricParams, callback) {
+    assert.object(client, 'napi client');
+    assert.UUID(userUuid, 'user uuid');
```
melloc: You want `assert.uuid` here (lowercase).

test/integration/helpers.js
```diff
@@ -1637,6 +1652,52 @@ function didRestifyHandlerRun(reqId, handlerName, callback) {
...
+    );
+}
+
+/*
+ * Gets or creates a fabric network for use in testing; based on the
+ * network *name*.
+ */
+function getOrCreateFabricNetwork(client, userUuid, vlan_id, params, callback) {
+    assert.object(client, 'napi client');
+    assert.UUID(userUuid, 'user uuid');
```
melloc: You want `assert.uuid` here (lowercase).

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,316 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+                    // this must be a prefix of the first half of the input
+                    var secondHalf = query.substr(32);
+                    var firstHalf = query.substr(0, 31);
+
+                    if (secondHalf.length >= firstHalf.length
+                        || secondHalf !==
+                        firstHalf.substr(0, secondHalf.length)) {
+
+                        log.info({ query: query },
+                            'Networks: impossible network id %s, skipping');
```
melloc: A third argument is needed to `log.info`, to replace the `%s`.

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,316 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+            }
+            return acc;
+        }, { err: null, result: null });
 
-        if (!external_net) {
-            callback(new errors.DockerError(
-                'unable to find "'+netName+'" network uuid'));
+        if (bestMatch.err) {
+            // found an error before a result.
+            log.error({ err: err, query: query },
```
melloc: Replace `err` with `bestMatch.err`?

lib/backends/sdc/containers.js
```diff
@@ -559,60 +525,316 @@ function addRulesToFWAPI(opts, rules, payload, callback) {
...
+                format('Networks: Error finding network to match %s', query));
+            callback(err);
```
melloc: And here.

lib/backends/sdc/containers.js
notmatt: I can't actually find anything that actually checks this on load, but the sapi template would appear to be incorrect without it. Strikes me that apart from asserting on this, we should probably add validation to config on startup.

rmustacc: When the `sdc_nat_pool` was added in NET-270 it was marked as optional; however, I think that the way we've been doing things, it's been treated as required in sdcadm. Likely some additional follow up work should be done to not assume it'll be there.

lib/backends/sdc/containers.js
notmatt: I think the best place to resolve this is with an assert on the query being non-empty; in the calling function we should never call `namedNetwork()` with an empty `HostConfig.NetworkMode`, so this should make it explicit.

lib/validate.js
```diff
@@ -392,6 +438,16 @@ function validateCreateContainer(req, res, next) {
...
+        assertOptionalString(container.HostConfig.NetworkMode,
+            'HostConfig.NetworkMode');
+
+        if (container.HostConfig.NetworkMode) {
+            assertNetworkMode(container.HostConfig.NetworkMode,
+                'HostConfig.NetworkMode');
+        }
+
+        // ensures mutually-compatible networking params.
+        assertNetworkingParams(container.HostConfig);
```
misterdjules: If I understand correctly, `assertNetworkParams` returns an error object when networking params are invalid, but this call-site doesn't check for the return value.

trentm: Ah, so `assertNetworkParams` should throw instead of return.

twhiteman:
This was already committed/merged a long time ago.

PR #118, opened 2017-05-11
DOCKER-1021 fix to return error when volume storage VM provision fails

joshwilsdon:


joshwilsdon:
Manual testing shows this working, still running test suite.

lib/backends/sdc/volumes.js
```diff
@@ -222,7 +244,17 @@ function deleteVolume(params, options, callback) {
                 }
 
                 pollVolumeState(volumeUuid, volapiClient, ['deleted', 'failed'],
-                    next);
+                    function _onPollComplete(err, vol) {
+                        var deleteErr = err;
+
+                        if (!err && vol.state === 'failed') {
```
misterdjules: Check that `vol` is not undefined with `if (!err && vol && vol.state === 'failed')`, or maybe assert that `vol` is not undefined when `err` is falsy.

lib/backends/sdc/volumes.js
```diff
@@ -109,7 +119,19 @@ function createVolume(volumeParams, options, callback) {
 
                 ctx.volume = volume;
                 pollVolumeState(volume.uuid, volapiClient,
-                    ['ready', 'failed'], done);
+                    ['ready', 'failed'],
+                    function _onPollComplete(err, vol) {
+                        var createErr = err;
+
+                        if (!err && vol.state === 'failed') {
```
misterdjules: Check that vol is not undefined with if (!err && vol && vol.state === 'failed'), or maybe assert that vol is not undefined when err is falsy.

misterdjules:
Added two minor comments, otherwise changes look good to me. Will review the tests when they're added. Thank you!

misterdjules:
@joshwilsdon Looks good to me, do you plan to include tests in this PR too?

test/integration/cli-nfs-shared-volumes-create-failure.test.js
```diff
@@ -0,0 +1,149 @@
...
+/*
+ * Copyright (c) 2017, Joyent, Inc.
+ */
+
+//
+// Summary:
+//
+// These tests ensure that when provisioning fails for a volapi storage VM, we
+// get an error message from sdc-docker rather than a message telling us the
+// creation was successful. We "break" provisioning for the 10g package by
```
misterdjules: nit: Maybe mention the ticket ID?

misterdjules: Just thinking out loud, not something I think we can fix now: I don't like the fact that this changes the ability to provision volumes while the test is running, but it doesn't really make things worse now. It would be a problem if e.g we ran some tests in parallel with other tests.

I have in my todo list to "create ticket for sdc-vmapi to be able to pass a flag that makes a provisioning workflow fail", but then we'd need to pass that flag from sdc-docker, not sure how sound that would be.

But again, this is just thinking out loud, nothing we can act on now, and I know you're already aware of this.

test/integration/cli-nfs-shared-volumes-create-failure.test.js
```diff
@@ -0,0 +1,149 @@
...
+                var results = [];
+                t.ifErr(err, 'get PAPI package');
+                if (pkgs) {
+                    results = pkgs.map(function mapUuids(pkg) {
+                        return pkg.uuid;
+                    });
+                }
+                t.equal(count, 1, 'should be 1 result ' + JSON.stringify(results));
+                if (count === 1 && results.length === 1) {
+                    PAPI_PACKAGE = results[0];
```
misterdjules: Can we just use `pkgs[0].uuid` instead and get rid of the `pkgs.map` above since we have to get `pkgs[0].traits anyway`?

test/integration/cli-nfs-shared-volumes-create-failure.test.js
```diff
@@ -0,0 +1,149 @@
...
+        + 'appropriate error message',
+        function (t) {
+            volumesCli.createTestVolume(ALICE_USER, {
+                size: '10g',
+                name: testVolumeName
+            }, function volumeCreated(err, stdout, stderr) {
+                var expectedErr = 'Error response from daemon: (InternalError) '
+                    + 'volume creation failed';
+                var matches;
+                var re = new RegExp(expectedErr.replace(/[()]/g, '\\$&'));
```
misterdjules: What does that do?

misterdjules:
@joshwilsdon Left a few comments on the latest changes.

test/integration/cli-nfs-shared-volumes-create-failure.test.js
joshwilsdon: Adding the ticket is a good idea. I'll do that.

And yeah, I don't have a better way atm for ensuring the provision fails. The runtest(s) scripts do ensure that we've indicated that this system is not a production system at least.

test/integration/cli-nfs-shared-volumes-create-failure.test.js
joshwilsdon: The reason for doing this was so that when the test failed, we only had the uuids in the assert message, so we can go look at which packages we got that we didn't expect. I'll change to just do the map there I guess.

test/integration/cli-nfs-shared-volumes-create-failure.test.js
joshwilsdon: I'll add a comment!

joshwilsdon:
🎉 🥇

misterdjules:
Thank you!

PR #121, opened 2017-06-09
DOCKER-1030 restrain number of size unit suffixes supported when crea…

joshwilsdon:
…ting NFS volumes


joshwilsdon:
merged with e3d3ce9db69a9a03055b22aae145f88ee8c0356b

PR #125, opened 2017-07-10
DOCKER-1034 support docker volume ls -f dangling=true

misterdjules:


lib/backends/sdc/volumes.js
```diff
@@ -153,15 +159,78 @@ function listVolumes(params, options, callback) {
...
+    var filters;
+    var filterValue;
     var log = options.log;
+    var reqParams = {};
+    var valIdx;
+
+    /*
+     * By default, we only list volumes that are ready. As far as I know, docker
+     * doesn't have a concept of a volume's state, and so doesn't provide any
+     * ability to filter on such a property.
```
misterdjules: That's a great question. I think it's ok to leave it as is in this PR since it's not a change from the previous behavior, but it seems it would be a good candidate for a separate ticket and investigation. Thoughts?

lib/backends/sdc/volumes.js
```diff
@@ -153,15 +159,78 @@ function listVolumes(params, options, callback) {
...
+        log.info({
+            filters: filters
+        }, 'filters');
+
+        for (filterName in filters) {
+            var filterValues = filters[filterName];
+            for (valIdx in filterValues) {
+                filterValue = filterValues[valIdx];
+                if (filterName === 'dangling') {
+                    filterValue = filterValue === 'true' ? true : false;
```
misterdjules: I agree, will change!

lib/backends/sdc/volumes.js
joshwilsdon: sounds fine to me!

misterdjules: Actually, I misread your question: with sdc-docker `docker volume create` returns only when the volume is in state `ready`, so at that point `docker volume ls` would include the volume that was just created in its output. Does that answer your question?

joshwilsdon: ah, even better.

misterdjules:
Merged with 9ff84449426add3777ff1fa56d4bfe1b5d17e16f.

PR #143, opened 2019-11-22
DOCKER-1110 stop docker container lookups from fetching all vms

twhiteman:


twhiteman:
Small update - just noticed docker only allows matching to the full container name - so we don't need the '*' matching against alias.

PR #146, opened 2020-02-11
TRITON-2065 Triton components should use the 20181206 min_platform

timfoster:


ad-m:
Why?

timfoster:
It's so that we can move components onto pkgsrc 2019Q4 base images. Since pkgsrc moved its min_platform, anything that uses that image must also be built on that new min_platform.

While _this_ component isn't yet planned to move over to that new min_platform, we decided it was saner to have our entire build environment move off its current 2015-era build machine onto something slightly newer rather than having different components requiring different build environments. Likewise, for users, not having to worry whether your platform image is modern enough to run a given component was something we wanted.

There's more detail in the following:

https://smartos.org/bugview/TRITON-2043
https://smartos.org/bugview/TRITON-1996

PR #148, opened 2020-03-30
TRITON-2096 Add affinity support to volapi

twhiteman:


lib/backends/sdc/containers.js
```diff
@@ -1974,7 +1974,7 @@ function buildVmPayload(opts, container, callback) {
                     cb(err);
                     return;
                 }
-                if (affinity) {
+                if (affinities) {
```
kusor: Good catch!

lib/backends/sdc/volumes.js
```diff
@@ -133,6 +135,30 @@ function createVolume(volumeParams, options, callback) {
                     });
             }
         },
+        function parseAffinityHints(_, cb) {
+            if (!payload.labels) {
+                cb();
+                return;
+            }
+
+            // This sucks, as I'd prefer DAPI to do this affinity parsing work
```
kusor: Even if it does, I'd personally go for better wording. Might be kids looking at this in the future

twhiteman: Heh - updated for the kiddo's.

PR #168, opened 2022-03-31
TRITON-2281 update sdc-docker base image to triton-origin-x86_64-21.4.0

bahamat:


danmcd (approved):
Found all of the TRITON issues with the results.  Thanks!

PR #170, opened 2022-09-16
TRITON-2327 Update sshpk to 1.13.2 for CVE-2018-3737

travispaul:


bahamat (changes requested):
Also please run the test suite (if there is one...I don't remember).

bahamat:
Ah, I see the test results in TRITON-2327, so good on that. Just need the URL fixed in package.json.

bahamat (approved):

